PR #3, opened 2025-02-13
feat: support both regex and non regex titles

saade:
Add support for both regex and non regex matching of titles.

The existing `toHaveTitle` expectation requires an string that will always be considered and compiled as regex, which can be limiting or prone to errors in cases where we need to match the exact title of a page.

PS: you may want to port that `Str::isRegex` to `Pest\Support\Str` namespace.


ijpatricio:
Solid!

PR #8, opened 2025-02-13
feat: add toNotHaveTitle operation

Jamesclark32:
Implements the [`.not` property option](https://playwright.dev/docs/api/class-pageassertions#page-assertions-not) on the `toHaveTitle` method as a distinct Operation class.


cndrsdrmn:
duplicate #1

PR #27, opened 2025-02-16
feat: Use html fixtures over real urls

jackbayliss:
## What it adds:
- This adds fixtures over using real urls  (we can pass file::// into the url so it loads as if it's a domain) 

## Why:
- If the website url tests we are using fails to load,  or even changes, it could introduce failing tests even though the tests are valid.
- Using fixtures should increase speed 🕺🏻 
- With adding more and more tests, we can ensure our expected behaviour is doing what we actually intend without relying on an external site 

Feel free to change it, i.e. the method names or whatever you see best. But I think it's a worthwhile addition. Or just lemme know what to change and I'll sort it when I get chance 🫡  give me a nudge and happy to implement on any tests that have been added to main

### DIFF
--- a/tests/Pest.php
+++ b/tests/Pest.php
@@ -20,3 +20,17 @@ function cleanupScreenshots(): void
 
     file_exists($basePath) && rmdir($basePath);
 }
+
+function htmlfixture($filename): string
+{
+    $file = __DIR__.'/Fixtures/html/'.$filename.'.html';
+
+    // We do so it works with Playwright correctly.
+    $file = str_replace(DIRECTORY_SEPARATOR, '/', $file);
+
+    if (! file_exists($file)) {
+        throw new RuntimeException("File '{$filename}' not found in fixture path: ({$file}");
+    }
+
+    return 'file://'.$file;
+}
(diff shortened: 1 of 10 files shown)


tests/Pest.php
```diff
@@ -20,3 +20,17 @@ function cleanupScreenshots(): void
 
     file_exists($basePath) && rmdir($basePath);
 }
+
+function htmlfixture($filename): string
+{
+    $file = __DIR__.'/Fixtures/html/'.$filename.'.html';
+
+    // We do so it works with Playwright correctly.
+    $file = str_replace(DIRECTORY_SEPARATOR, '/', $file);
```
jackbayliss, line 29: This makes it work for Windows, feel free to change it though

clemblanco:
Pretty sure the idea was to use #11 in order to run tests on a real Laravel app running from the `/playground` folder instead of using fixtures. This would allow us to test proper integrations with Laravel with for example authentication.

jackbayliss:
@clemblanco  ah cool, fair enough my bad didn't realise 😂

PR #28, opened 2025-02-16
chore - Add cache for node modules / playwright

jackbayliss:
Speeds up tests ,WIP


jackbayliss:
I think we're already caching it, and doesn't seem to improve that much IMO 👋🏻  Time for sleep.

PR #31, opened 2025-02-16
Add AssertMatchesScreenshot & AssertNotMatchesScreenshot operations

vaishnavyogesh:
Changes:
- Added `Operations\AssertMatchesScreenshot` to check whether the page's current screenshot matches with stored screenshot
- Added `Operations\AssertNotMatchesScreenshot` operation to check whether the page's current screenshot doesn't match with stored screenshot

Used `toHaveScreenshot()` api instead of `toMatchScreenshot()` as suggested [here](https://playwright.dev/docs/api/class-snapshotassertions)

### DIFF
--- a/src/Operations/AssertMatchesScreenshot.php
+++ b/src/Operations/AssertMatchesScreenshot.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Operations;
+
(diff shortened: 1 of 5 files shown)


ijpatricio:
We're changing architecture for Playwright Worker + Websocket client to send instructions.
Just a heads up. Your PR is getting merged, but the code will be changed for the new architecture soon.

Thanks for your contribution!

PR #33, opened 2025-02-16
feat: Add `when` operation for conditional actions

AUnhurian:
I have implemented the `when` operation for tests with the ability to set `then` and `else` callbacks.
This will provide more flexibility for customers when using it.

I would be happy to receive any comments or recommendations.

Example:

**With both callbacks (`then` and `else`):**
```PHP
$url = 'https://laravel.com';

$browser = $this->visit($url);

$browser->when(
    new Pest\Browser\Conditions\See('Laravel - The PHP Framework For Web Artisans'),
    function (Pest\Browser\PendingTest $browser): void {
        $browser->clickLink('Get Started')
            ->assertSee('Installation');
    },
    function (Pest\Browser\PendingTest $browser): void {
        $browser->assertSee('Laravel');
    }
);
```

**Only `then` callback:**

```PHP
$url = 'https://laravel.com';

$browser = $this->visit($url);

$browser->when(
    new Pest\Browser\Conditions\See('Laravel - The PHP Framework For Web Artisans'),
    function (Pest\Browser\PendingTest $browser): void {
        $browser->clickLink('Get Started')
            ->assertSee('Installation');
    }
);
```

_Compile result:_
```JS
import { test, expect } from '@playwright/test';

test('runtime', async ({ page }) => {
	await page.goto('https://laravel.com');
	if (((await page.locator('body').textContent()).includes("Laravel - The PHP Framework For Web Artisans"))) {
		await page.locator('a').filter({ hasText: /Get Started/i }).click();
		await expect(page.locator('body')).toHaveText(/Installation/);
        }
});
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -100,6 +100,7 @@ TBD
 - [pause](#pause)
 - [refresh](#refresh)
 - [rightClick](#rightClick)
+- [when](#when)
 - [screenshot](#screenshot)
 - [visit](#visit)
 
@@ -225,6 +226,23 @@ Takes a full-page screenshot of the current page and saves it under `/Browser/sc
 
 ```php
 $this->screenshot('filename');
+```
+
+### when
+
+The when operation in Pest Browser allows you to execute different actions based on whether a specific condition is met when visiting a webpage. This feature provides dynamic test execution, enhancing the flexibility of browser tests.
+
+```php
+$this->when(
+        new Pest\Browser\Conditions\See('Laravel - The PHP Framework For Web Artisans'),
+        function (Pest\Browser\PendingTest $browser): void {
+            $browser->clickLink('Get Started')
+                ->assertSee('Installation');
+        },
+        function (Pest\Browser\PendingTest $browser): void {
+            $browser->assertSee('Laravel');
+        }
+    );
 ```
 
 ### visit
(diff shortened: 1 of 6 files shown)


AUnhurian:
It may be a draft, but I believe that together we can make it better =)

ju5t:
I think there's potential for a lot of overlap between assertions and the introduced conditions. I don't think there's a reason why you couldn't use an operation for it; apart from the naming of operations being a bit confusing when used in conditionals.

I do think it can be an interesting feature. It has some similarities with `whenAvailable` in Dusk.

AUnhurian:
Thanks, I agree. I can change the method's name to `whenAvailable`

ju5t:
@AUnhurian `whenAvailable` is limited to a selector. I think having your current PR under the hood is a great option where `whenAvailable` can build upon. But it's up to @nunomaduro to decide on this.

AUnhurian:
Okay, I will wait for the final decision on what changes are necessary

ijpatricio:
We're changing architecture for Playwright Worker + Websocket client to send instructions.
Just a heads up. Your PR is getting merged, but the code will be changed for the new architecture soon.

Thanks for your contribution!

PR #35, opened 2025-02-16
Add assertVisible and assertMissing assertions

dzmitry-vasileuski:
Add assertVisible (https://laravel.com/docs/11.x/dusk#assert-visible) and assertMissing (https://laravel.com/docs/11.x/dusk#assert-missing) assertions.

⚠️I used Dusk assertion names. But I don't think `missing` is a good name for such assertion. Would be better to have `assertHidden` instead. Probably, we can add this method to the PendingTest as an alias to `assertMissing`.


ijpatricio:
Woohoo, thanks @dzmitry-vasileuski

PR #36, opened 2025-02-16
chore: setup playground

clemblanco:
- automatically configure the Laravel app in `/playground` upon `composer install/update` so we don't have to maintain and manually call a `/playground/install` script
- configure `playground/composer.json` to automatically create the SQLite database if it doesn't exist and migrate it upon `composer install`
- renamed `welcome` view from the playground app into `index`
- remove unnecessary `playground/.env.example` file
- configure the `webServer` within `playwright.config.js` to automatically start the Laravel app from `/playground` before running some tests


clemblanco:
Hey @nunomaduro thanks man.

Just wanted to mention that I usually "squash" instead of "merge" and use the PR description as the commit message so I don't mind about WIP commits on PRs as they will be squashed into one... 

Let me know if you think you prefer to "merge" instead of "squashing" and I'll try to avoid the WIP commits.

PR #37, opened 2025-02-16
Add click operation

dzmitry-vasileuski:
Add click operation (https://laravel.com/docs/11.x/dusk#clicking-on-elements).


ijpatricio:
Thank you @dzmitry-vasileuski

PR #38, opened 2025-02-16
feat : Assert attribute

jackbayliss:
Adds ability to assert an attribute

 for now, it checks the attribute on a selector matches an exact value, which is what dusk does 😉 

Let me know what you think 🫡


ijpatricio:
Hey @jackbayliss 

Thank you for you PR.

Follow Dusk API, everything looks well!

Can you please add the relevant documentation part, in the README.md?

There's already example there, following Dusk's guidelines

```
### Available Assertions

<a name="assert-dont-see"></a>
#### assertDontSee

Assert that the given text is not present on the page:

```php
test('assert does not see', function () {
    $url = 'https://laravel.com';

    $this->visit($url)
        ->assertDontSee('we are a streaming service');
});
```
```

jackbayliss:
Sure 😁 will adjust shortly 💜

jackbayliss:
Should hopefully be better now, I also fixed the markdown so it displays Available Assertions properly.  

Also removed the test function from the snippets 🫡  

Let me know if you need anything else.

ijpatricio:
Looks good to me! Thank you! 👍 

If you want to keep the steam up, there [assertAttributeMissing](https://laravel.com/docs/11.x/dusk#assert-attribute-missing) next :) :)

PR #40, opened 2025-02-16
Add assertScript assertion

dzmitry-vasileuski:
Add assertScript assertion (https://laravel.com/docs/11.x/dusk#assert-script)


ijpatricio:
Again, this looks very good. Thank you!

Can you please add the docs, like mentioned before?

ijpatricio:
Thank you, @dzmitry-vasileuski !

PR #41, opened 2025-02-16
chore: make the tests use the playground instead of laravel.com

clemblanco:
- ~coded~ copied a simple naive playground index page. We can add whatever we need on that page to test various interactions (double click, textareas...)
- migrate the existing tests to use the playground http://localhost:9357 instead of https://laravel.com
- making playground `host`+`port` static for now `http://localhost:9357` as we need to assert against it in some tests
- ignoring `package-lock.json` from `/playground` as we don't need lock files in the playground
- automatically running `npm install` upon `composer install` for the `/playground` so that we automatically install npm dependencies for the playground too
- automatically running `npm run build` upon `npm install` for the `/playground` so that we automatically build the assets whenever the playground gets installed

image


ijpatricio:
Hello @clemblanco !

All seems very good, nice job!

I just have one question, which is why would we ignore package-lock.json ? The playground is an application we maintain just as any consumer project, and not haveing version controlo on PHP and JS packages may lead into tests breaking.

ijpatricio:
@clemblanco I figure better to go ahead and merge, there's a lot of PR to go in, and playground is missing.

I made some changes:
- Keeping package.lock in the playground, reason explained above.
- In playwright config, keeping the reused variables defined in one place.
- Using `npm ci` (not install, that would change package-lock all the time)
- Instead of `npm dev` calling the `build`, we do both from composer.json - more explicit on the source.

ijpatricio:
Thank you @clemblanco , great job!! 🚀

clemblanco:
Hey @ijpatricio I think there are two different things:
1. `composer.lock` + `package-lock.json` for the plugin itself
2. `composer.lock` + `package-lock.json` for the playground app only

The plugin already ignores them because it's a plugin and we obviously rely on the consuming app to pin down and lock specific versions so the only constraints we enforce are the major/minor non-breaking versions, meaning upon installation we will let the consuming app install whatever latest version is available according to the major/minor specs we have for the plugin.

Thinking about it, for the playground app, I don't see why we couldn't do the same as we would actually want to know if something is breaking in the pipeline according to new releases of whatever the playground app is using. Yes for a Laravel app we wouldn't want this normally so things are locked properly but here I think it would be okay as it would highlight a breaking change without us having to upgrade anything manually.

I was looking at how https://github.com/orchestral/testbench and I don't think they pin down any specific versions for the dummy Laravel app they use either.

Happy with whatever tho :)

ijpatricio:
Hmm, it seems you reinforced the point of lock the playground.

To make things materialized: if we install Livewire X, React Y, Vue Z, we would want to lock those versions, right?
We desire to have an usable app, where we have a known state, and therefore, we can test the package behavior.
This is why goes in favor of locking, IMO.

> Thinking about it, for the playground app, I don't see why we couldn't do the same as we would actually want to know if something is breaking in the pipeline according to new releases of whatever the playground app is using

Why are we interested in failing a package behaviour test when say Vue updates something in the app? That's out of scope.

I might be missing something, no doubt! I just didn't see it yet.

What goes in favor of not locking, would be to leave the Playwright package and browsers unlocked. And also specifically target to support N amount of latest versions?

Does this clear things up?

clemblanco:
Haha not really sorry.

If we already have something like `^1.2.3` in a non-lock file I think it should be good enough for a playground app. But it's just my personal opinion and it's probably not that important, just means more maintenance to update the lock files too.

PR #42, opened 2025-02-16
feat: Add `pause` operation to PendingTest

AUnhurian:
This is a pull request that adds a method similar to Laravel Dusk's `pause` (https://laravel.com/docs/11.x/dusk#waiting).

It will pause for a specified number of milliseconds, which the user can set in the method (default is set to 1000 ms).

Example:

```PHP
$this->visit('https://laravel.com')
        ->clickLink('Get Started')
        ->pause(5000)
        ->assertUrlIs('https://laravel.com/docs/11.x');
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,3 +62,21 @@ test('assert does not see', function () {
         ->assertDontSee('we are a streaming service');
 });
 ```
+
+### Available Operations
+
+#### pause
+
+It will pause the process for the specified number of milliseconds you provide.
+
+Pause the test execution for 5 seconds:
+
+```php
+test('pause', function () {
+    $url = 'https://laravel.com';
+
+    $this->visit($url)
+        ->pause(5000)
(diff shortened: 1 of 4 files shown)


ijpatricio:
Hello @AUnhurian 

Thank you for your PR!

May you please change the implementation to  use the `waitForTimeout` API in PW? https://playwright.dev/docs/api/class-page#page-wait-for-timeout

Also, can you please add the relevant documentation part, in the README.md?

There's already example there, following Dusk's guidelines

```
### Available Assertions

<a name="assert-dont-see"></a>
#### assertDontSee

Assert that the given text is not present on the page:

```php
test('assert does not see', function () {
    $url = 'https://laravel.com';

    $this->visit($url)
        ->assertDontSee('we are a streaming service');
});
```
```

ijpatricio:
Thank you @AUnhurian 🚀

PR #46, opened 2025-02-17
Update README.md

parijke:
To prevent that you have to explain it in every livestream :-) (with i little help of AI for the playwright stuff)


nunomaduro:
Thanks; but rather not elaborate on this here.

PR #47, opened 2025-02-17
chore(ci): only install chromium driver to speed things up

clemblanco:
- as this is the only driver we currently use...


ijpatricio:
Yeah, at least while we only use Chrome. Thanks!

PR #51, opened 2025-02-17
fix: assertSee / assertDontSee special characters

clemblanco:
Given the HTML `<span>foo (bar) baz</span>`, using `assertSee("foo (bar) baz")` would understand `(` and `)` as regex special characters and compile it into
```js
wait expect(page.locator('body')).toHaveText(/foo (bar) baz/i);
```
We would like instead to escape such regex special characters so it would become
```js
wait expect(page.locator('body')).toContainText("foo (bar) baz", { ignoreCase: true });
```

Switching from `toHaveText()` for `toContainText()` without using Regex to be closer to Dusk's current implementation as it only relies on `Str::contains("text", ignoreCase: true)` currently.

The difference between `toHaveText()` and `toContainText()` will be around how it handles HTML strings. If you have the following HTML
```html
Pest Plugin Browser<br />Playground
```
`toHaveText("Pest Plugin Playground")` will be `false` as it will remove HTML tags during the string comparison. It will try to find
```
Pest Plugin Browser
```
within
```
Pest Plugin BrowserPlayground
```
as it uses a "match whole word" approach. Notice the `<br />` being removed.

However, with `toContainText()`, it will indeed find `Pest Plugin Browser` within `Pest Plugin BrowserPlayground`, just like `Str::contains()` would.


ijpatricio:
So this topic is because of the api that accepts both text and regex, no??

Almost sure this was well intentioned back then, to be able to handle both, but it raised this edge cases.

If Dusk isn't accepting both text and regex, I believe we can safely go ahead and allow only text. Having an API for regex seems preferred.

clemblanco:
So Dusk simply uses https://laravel.com/docs/11.x/strings#method-str-contains by default. No regex. However it can be case sensitive and insensitive.

I believe if we really want to stick to how Dusk does it we should use
```js
expect(page.locator('body')).toHaveText('text', { ignoreCase: false }); // or true
```
Instead of
```js
expect(page.locator('body')).toHaveText(/text/); // or /text/i
```

See https://github.com/microsoft/playwright/issues/13927

I'll update the PR.

ijpatricio:
Yeah, agree! Regex has to be in a different API

PR #52, opened 2025-02-17
refactor: `assertDontSee` should ignore case by default

clemblanco:
I don't think it makes sense for `assertDontSee()` to be case sensitive by default. Even if it's what Dusk uses currently.

Usually when you're making sure some text is not on the page, you want to make sure that for example `Foo Bar` or even `foo bar` are not present. It feels like it could lead more to a false-positive in a test suite if we use `assertDontSee("Foo Bar")` and the HTML has `foo bar` but the test is green.

To me, it makes more sense for it to be case insensitive by default, which is the opposite of how `assertSee()` works.

I might be wrong, feel free to ignore this.

### DIFF
--- a/src/Operations/AssertDontSee.php
+++ b/src/Operations/AssertDontSee.php
@@ -13,7 +13,7 @@ final readonly class AssertDontSee implements Operation
      */
     public function __construct(
         private string $text,
-        private bool $ignoreCase = false,
+        private bool $ignoreCase = true,
     ) {
         //
     }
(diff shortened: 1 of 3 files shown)


ju5t:
Laravel/PHPUnit is usually stricter by default and lenient if you ask it to. For example: 

- PHPUnit has an `assertEquals` assertion. This is case sensitive.
- assertStringContainsString is case sensitive.
- `Str::contains()` is case sensitive.

And there's probably more.

So having it case sensitive by default should make sense for most. And I think if we're following Dusk's method names we should also follow the method signature/behavior.

clemblanco:
Hum yeah it's probably just me then.

PR #55, opened 2025-02-17
feat: process assertion failure

dzmitry-vasileuski:
Process assertion failure to see in console what exactly fails in our browser test.

The logic is:
1. Take the line of playwright test where the error happened
2. Find failed assertion by comparing its compiled form with the given line
3. Throw an exception in the failed assertion

Output example:
Screenshot 2025-02-17 192925


clemblanco:
Could we see the line where we call `visit()` and it failed in the test instead of where the Exception is thrown? I have no idea, just asking how feasible it is.

PR #62, opened 2025-02-17
fix : Tests - php stan / rector

jackbayliss:
PHP stan & Rector are both failing:

 
[Stan](https://github.com/pestphp/pest-plugin-browser/actions/runs/13378086137/job/37361440177)
[Rector](https://github.com/pestphp/pest-plugin-browser/actions/runs/13379562533/job/37365595896)

This PR fixes it and Rector & Stan are both happy 👯 

I believe https://github.com/pestphp/pest-plugin-browser/pull/40 introduced it

PR #63, opened 2025-02-17
chore / fix : url changes to use playground

jackbayliss:
All tests now use the playground (localhost) rather than Laravel.com


jackbayliss:
Failing due to rector / stan which is fixed by https://github.com/pestphp/pest-plugin-browser/pull/62 

Marking as ready, as it's passing tests - would suggest to merge ^ first

clemblanco:
Thanks! I think I've migrated the existing tests to the playground in this PR already https://github.com/pestphp/pest-plugin-browser/pull/48 but should have probably split it in a separate PR tbh.

Let's consolidate 💪

PR #65, opened 2025-02-18
feat: added new operations - Check, UnCheck & AssertChecked, AssertNo…

jleontidis:
Adds the following operations :
  - assertChecked and assertNotChecked assertions (https://laravel.com/docs/11.x/dusk#assert-checked)
  - check and uncheck method (https://laravel.com/docs/11.x/dusk#checkboxes)
  - documentation
  - added checkboxes to playground
  - ignored uncheck word in peck.json


ju5t:
It's better to disable any prettier/linting plugins in your editor as they often result in large diffs, like they did here. That's not to say they can't be used, but linting should only be enforced by the package.

Perhaps we should also decide how much styling is necessary when adding new content to the playground. Personally I think it should be as simple as possible with as little styling as possible. Only add what's necessary.

jleontidis:
You are right. 
It is the automatic tailwind class sorting with Prettier that changed the order of all classes in the playground.
Should i disable it and resubmit the PR ?

ijpatricio:
@jleontidis thank you for your work! yes please!

jleontidis:
I updated the PR.

PR #78, opened 2025-02-19
Remove duplicate method. Fix coordinates on Point and CLick test.

ijpatricio:


clemblanco:
Hum... 🤔 Was green on the PR you just merged. Do you want me to look at it?

ijpatricio:
Hey @clemblanco 

Thanks for promptly help!! 💪 


No, dupe was that I didn't want to not include the duplicated work, so I did, and ended up with that. No worries.

Yeah, I'm struggling with 2 things:

- The coordinates: in my machine, it fails for that coordinates (oh no, "on my machine"). I'm almost sure the default resolution is not being applied. I checked while running the rendered `runtime.spec.js` with --debug. Playwright themselves disencourage the use of this, in favor of using a locator. Probaly better to make a page with like a small "chess table" matrix of squares, when clicked the show coordinate A1, A2, etc....

- The click and hold test seems flaky on this last test run. The component has a step Interval of 100ms, meaning a tiny delay and click 1 second may not trigger the action. Better to make the default click and hold for 1500ms? or reduce the component for 95ms?
image
image

The report could also be in html, generate the file, and we'll have GH actions upload the Artifact so we can open it, what do you think?

I might be wrong about an of this! Yes, help is appreciated! I have to go dinner, wife will kill me soon if I don't show up :)

clemblanco:
@ijpatricio this #79 should do, I've explained things on there.

ijpatricio:
Thanks @clemblanco  merged!

Running tests locally as well.

CleanShot 2025-02-19 at 23 27 58@2x

Any idea why this fails?

I'm guessing it's because the resizing is not happening same in different environments...

ijpatricio:
If I run PW just for that test, I can find good coordinates, and that would work locally.

But then in CI it won't behave in the same fashion.

ijpatricio:
But yeah, the ClickAndHold seems fixed! Let's move to another issue

clemblanco:
Did you manage to get `clickAtPoint` failing locally? I will run it locally another 100 times and see...

PR #79, opened 2025-02-19
chore(ci): fix pipeline

clemblanco:
- making it green and investigating into potential flakiness

1. `ClickAndHoldTest()`

So this is sometimes failing because both the test dummy button and the `clickAndHold()` function were both using exactly `1000ms`. Sometimes one could be off by one micro-second and the button wouldn't be hold long enough.

image

I fixed this by setting up the test dummy button on `~700ms` instead so `clickAndHold()` would always have enough time.

Ran the test 100 times consecutively locally and it's now always green.

image

2. `clickAtPoint()`

I think the coordinates were good in the end and it's been green 100 times for me... 🤔 

image

### DIFF
--- a/playground/resources/views/livewire/button.blade.php
+++ b/playground/resources/views/livewire/button.blade.php
@@ -6,7 +6,7 @@
             x-on:{{ $event }}="$wire.handle"
         @endforeach
         @if ($flavor === 'hold')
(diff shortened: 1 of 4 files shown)


ijpatricio:
Good one @clemblanco !

ijpatricio:
You are crushing it man 💪

PR #80, opened 2025-02-19
fix: html comment livewire playground

clemblanco:
- moving from html to blade comment (my bad)
- silence Livewire warning, see https://github.com/livewire/livewire/discussions/6464

image


ijpatricio:
Was exactly noticing this :) 

Thanks.

I'm trying to solve now the "Click at point/coordinates"

clemblanco:
Same, #81 seems fine, give it a go locally maybe.

PR #81, opened 2025-02-19
fix(ci): click at point flakiness

clemblanco:
- making sure we see the button before clicking it can help prevent the weird flakiness we seem to see

image


ijpatricio:
Hopefully, then!!

I was going with something like a matrix. the text could be hidden, and if we click at Cell, the text displays.

Ok, but if this fixes, that's good!


CleanShot 2025-02-19 at 23 51 20@2x

Issue #12, opened 2025-02-14
Understand playwrights worker API


nuernbergerA:
TDLR: WebSocket is the way to go

Other Languages
https://github.com/mxschmitt/awesome-playwright?tab=readme-ov-file#language-support

So far ruby has good illustrations
https://playwright-ruby-client.vercel.app/docs/article/guides/launch_browser

https://github.com/YusukeIwaki/playwright-ruby-client?tab=readme-ov-file#communicate-with-playwright-server

ijpatricio:
Adding my personal experience/findings:

I haven't done WS and PHP ever :) Should be fun!

Started tinkering with WS and PHP here: https://github.com/pestphp/pest-plugin-browser/blob/pw-server/runner.php
There's the composer package installed `ratchet/pawl` and the `runner.php` file. Just want to make a most basic thing first.

Basically, we need to start a `npx playwright run-server` and it gets running.
Then connect to it via WebSocket, by running the `./runner` - which is not getting anywhere yet!

There's no Documentation to be found, about Playwright and the Server API, on the searches I've done.
Theres the code to be explored in:

- `node_modules/playwright-core/lib/remote`
- `node_modules/playwright-core/lib/server`

Let us know your experience :)

ijpatricio:
Hey @formula23  - like we talked in #64 , the `pw-server` isn't very far along.

I didn't get it nowhere past the initial connection - no docs to be found, it seems.

Next thing I would do is to look at the Ruby client and try to follow/mimic the steps - https://playwright-ruby-client.vercel.app/docs/article/getting_started

Since it's a WebSocket protocol, we just have to adhere to the same. The Ruby Client and the node_modules dirs
- node_modules/playwright-core/lib/remote
- node_modules/playwright-core/lib/server

Should give a good head start. We need to dive in these, it seems

clemblanco:
Stupid question but could we transpose all the tests we wanna run from Pest into one single file like we currently do but instead of running it once per Pest test case, we wrap each test case with jest `it()` or `describe()` and collect the results afterwards? This way we run Playwright once only. Unsure if this would be simpler or not... Sorry if this is very stupid and/or naive.

ijpatricio:
There's no "stupid questions", they're just questions 😄 I like how you challenge the norm.

I do feel that the WebSocket is the robust option, when using PHP, because the API becomes JSON, and we can do things like

```php

test('assert visible', function () {

    // Action 1
[...]

PR #82, opened 2025-02-20
ci: exposing screenshot/video artifacts for failing tests

clemblanco:
- this way we can have access to the screenshot/video of the failing tests directly from GitHub Actions
- moving the output dir to `.temp` and adjusting `.gitignore` accordingly
- added bonus: this will also be accessible when running the test suite locally within the `.temp` directory


clemblanco:
Closing for now as we need #12 to be sorted before.

PR #87, opened 2025-02-22
feat: process failed assertions

dzmitry-vasileuski:
Added processing of failed assertion.
Now CLI shows related exception message and shows where failed assertion was called.

For now I left auto-generated exception message in the following format:
`Failed <assertion> assertion with <args>`
But we can pass more detailed message as a second argument of `Assertion` constuctor.

Used `debug_backtrace` function to find a file and line of called assertion. Didn't find a better way.

image

### DIFF
--- a/peck.json
+++ b/peck.json
@@ -7,7 +7,8 @@
             "str",
             "dont",
             "doesnt",
-            "uncheck"
+            "uncheck",
+            "args"
         ],
         "paths": []
     }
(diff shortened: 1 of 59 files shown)


ju5t:
Similar to my question in your previous PR: I assume any operation can fail, not just assertions? Why limit this to assertions?

ju5t:
I think @nunomaduro or @ijpatricio should decide on that.

An abstract `Operation` class feels like a cleaner solution; but it's not up to me :)

clemblanco:
Do we wanna forward outputs like this from Playwright?

image

Issue #21, opened 2025-02-15
Running the server


ijpatricio:
When we ask Playwright to run the server, it will run PHP process under the Node process.

Funny enough, such a Node process is being run by a PHP process - the original `./vendor/bin/pest`.

Because of sudden interruptions, some orphan processes might be left alive.

We can check by running, in MacOS
```
# check which processes are listening on port  9357
lsof -i:9357

# kill all of those processes
kill $(lsof -t -i:9357)
```

The problem will arise with other OSes as well, Windows and Linux.

So, before proceeding with more work, I'm asking for opinion.

Should we configure Playwright to not start webservers?
It seems better to me.
We are responsible to start them before executing `playwright test`.
We remove one more layer.

We should do it from PHP or Bash.

Are you onboard with this?

ju5t:
In the stream `pest()->beforeAll()` and `pest()->afterAll()` didn't work for some reason, but that feels like a good location to start the webserver. Similar to how it cleans up the screenshots that get made during tests.

ijpatricio:
@ju5t yeah, if we start that in PHP, we remove one layer.

The situation becomes: the Pest process say (pid 1010), starts one ore more `artisan serve`.
We then get pid 1010, 1011, 1012, 1013 (examples).


If the original 1010, suddenly dies/throw, it may not properly clean up 1011-1013.

Then, when you run pest again, it will fail to start artisans because such ports are already initialized.

Therefore, I think we need some manager, which handles the webservers, knows the ports, and instructs Pest as well (so if it dies, the manager still lives, and cleans up.

![Image](https://github.com/user-attachments/assets/30b15749-0c0f-4d3c-8388-9f684b004938)

PR #88, opened 2025-02-23
feat: start playground from php instead of playwright process

dzmitry-vasileuski:
Added the `Playground` class to run `php artisan serve` command from php instead of Playwright.

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,17 +46,13 @@ To run the test suite, execute:
 
 For each Operation/Assertion, we add a corresponding Test.
 
-We can make use of the `playgroundUrl()` helper, to get its URL during the test.
+We can make use of the `playground()->url()` helper, to get its URL during the test.
 
-We can provide a URI that will be appended, e.g: `playgroundUrl('/test/interactive-elements')`.
-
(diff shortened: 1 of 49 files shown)


ijpatricio:
Because we're changing architecture for Playwright Worker + Websocket client to send instructions.

So we will need to have some "Manager" controlling the underlying processes.

https://github.com/pestphp/pest-plugin-browser/issues/21#issuecomment-2661107426

I'm not sure how it's going to be, if PHP is enough to run those in parallel, or if something cross platform like Go would make a binary for that.

Merging for now, you had a tremendous effort in this move, and hopefully we'll keep things PHP.

I went through learning the bare naked things, and I feel we're ready to start moving on.

I checked with @nunomaduro and we're going with the current WS client. even being outdated, but it has a Sync implementation, and the risk is not huge.

What we have now is a mountain of files from the previous implementation, so I'm thinking in how to process that :)

The most sane way is to clean up, getting a starting point with minimal setup on the new architecture, and then by going back to earlier branch, and porting the functionality to the so called Wire Protocol.

Thanks for your hard work @dzmitry-vasileuski

PR #89, opened 2025-03-07
Feature/playwright websocket

dzmitry-vasileuski:


src/Playwright/Client.php
```diff
@@ -0,0 +1,89 @@
...
+            'method' => $method,
+            'params' => $params,
+            'metadata' => $meta,
+        ]);
+
+        // file_put_contents('log.log', $requestJson.PHP_EOL, FILE_APPEND);
+
+        $this->websocketClient->text($requestJson);
+
+        while (true) {
```
clemblanco, line 69: Bit weird to create our own infinite loop to keep receiving messages... but I guess this is similar to what you'd get with a ReactPHP loop.

src/Playwright/Element.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * Check element.
+     */
+    public function check(): void
+    {
+        $response = Client::instance()->execute($this->guid, 'check');
+
+        foreach ($response as $message) {
+            // read all messages to clear the response
+        }
```
clemblanco, line 62: Bit funny

src/Playwright/Frame.php
```diff
@@ -0,0 +1,147 @@
...
+            }
+        }
+
+        return $this;
+    }
+
+    /**
+     * Returns the meta title.
+     */
+    public function title(): string
```
clemblanco, line 49: Maybe `getTitle()` would be more appropriate here.

src/Playwright/Server.php
```diff
@@ -0,0 +1,114 @@
...
+
+    /**
+     * Returns the server instance.
+     */
+    public static function instance(): self
+    {
+        if (! self::$instance instanceof self) {
+            self::$instance = new self(
+                (string) ($_ENV['PEST_BROWSER_PLAYWRIGHT_HOST'] ?? self::DEFAULT_HOST),
+                (int) ($_ENV['PEST_BROWSER_PLAYWRIGHT_PORT'] ?? self::DEFAULT_PORT),
```
clemblanco, line 46: `PEST_BROWSER_(HOST|PORT)` is probably enough

src/Playwright/Server.php
```diff
@@ -0,0 +1,114 @@
...
+     * Starts the Playwright server.
+     */
+    public function start(): void
+    {
+        if ($this->isRunning()) {
+            return;
+        }
+
+        $this->process = new Process([
+            'npx',
```
clemblanco, line 63: I believe if we use `npx` it will automatically pull `playwright` before running the binary (if missing) which is why it wasn't in the `package.json` until now.

src/Playwright/Server.php
```diff
@@ -0,0 +1,114 @@
...
+            fn ($type, $output): bool => $output === "Listening on {$this->url()}\n"
+        );
+    }
+
+    /**
+     * Checks if the specified port is free for use.
+     */
+    public function isRunning(): bool
+    {
+        $process = Process::fromShellCommandline("lsof -t -i :{$this->port} 2>/dev/null");
```
clemblanco, line 84: Would `lsof` available in all environments?

src/Playwright/Server.php
```diff
@@ -0,0 +1,114 @@
...
+        return count($pids) > 0;
+    }
+
+    /**
+     * Stops the Playwright server.
+     */
+    public function stop(): void
+    {
+        if ($this->isRunning()) {
+            // Stop this way because initial process changes PID for some reason
```
clemblanco, line 100: Maybe instead of trying to be too smart by using `lsof` and `kill` we could start by being quite naive and just throw an error if something went wrong for now. I believe @ijpatricio mentioned something about ports in #21 too.

tests/Browser/Operations/AssertChecked.php
clemblanco, line 1: Feels like `Test` is missing at the end of the filename.

tests/Browser/Operations/AssertQueryStringMissingTest.php
```diff
@@ -2,14 +2,18 @@
...
-    $this->visit('/?q=test')
-        ->assertQueryStringHas('q')
-        ->assertQueryStringHas('q', 'test')
-        ->assertQueryStringMissing('q', 'test-1');
-});
+use PHPUnit\Framework\ExpectationFailedException;
+
+describe('assertQueryStringMissing', function () {
+    it('passes when the query string does not have the given query param', function () {
+        $this->visit(playgroundUrl('/?q=test'))
```
clemblanco, line 9: Maybe this could be simpler.
```suggestion
        $this->visit(playground('/?q=test'))
```
Unrelated to this PR anyway, ignore for now.

tests/Browser/Operations/ClickAndHoldTest.php.bk
clemblanco, line 1: Some backup files `.php.bk`?

In general it feels like the changes you've made in the existing tests are good but could probably be done in a separate PR, later on, as it is adding more changes to an already large PR without being absolutely necessary for the websocket implementation to work.

This PR is already quite of an overhaul of the underlying structure in terms of how things work using websockets. This is huge and this is some pretty amazing stuff but it would be nice if we could maybe isolate only this first move in this PR: shift in paradigm - before improving the test suite and working on other bits.

src/Autoload.php
```diff
@@ -14,6 +17,15 @@
...
+
         return (new PendingTest)->visit($url);
     }
 }
+
+register_shutdown_function(function () {
+    if (Parallel::isEnabled() || ! Parallel::isWorker()) {
+        Server::instance()->stop();
+    }
+});
```
clemblanco, line 31: Is this only to support parallel testing?

When would this be called exactly?

It could be good to have in the PR description a quick summary of the lifecycle of the test suite after this change maybe.

The paradigm shift here seems pretty amazing as most of it would become PHP-based, using websockets to "give orders" to the Playwright engine/server, as far as I understand, but I'm curious to understand the impact on how things get executed, with and without parallel testing.

src/Operations/AssertAttributeMissing.php
```diff
@@ -4,29 +4,24 @@
...
-        return sprintf(
-            'const attributeValue = await page.locator("%s").getAttribute("%s"); await expect(attributeValue).toBeNull();',
-            $this->selector,
-            $this->attribute
-        );
+        $value = $this->page->getAttribute($selector, $attribute);
+
+        expect($value)->toBeNull();
+
+        return $this;
```
clemblanco, line 25: All of our assertions and operations will become PHP-based which is pretty amazing and will solve the "error reporting" aspect I guess.

I'm curious to see if we can get some sort of helpful output from a failing assertion for example with something roughly looking like this
image

Would be good to include some screenshots of these sorts of things in the PR description maybe if it's working "as it is".

If it is not, it's fine too, but we just need to be aware that we will need to consolidate/work on the reporting aspect maybe at a later point.

src/Operations/AssertScript.php
clemblanco, line 1: Gone completely?

src/Operations/ControlClick.php
clemblanco, line 1: Gone?

src/Operations/ClickAtXPath.php
clemblanco, line 1: Gone?

src/Operations/ClickAtPoint.php
clemblanco, line 1: Gone?

src/Operations/ClickAndHold.php
clemblanco, line 1: Gone?

src/Operations/RightClick.php
clemblanco, line 1: Gone?

clemblanco (commented):
This is some pretty amazing stuff, everything can become PHP-based and that's awesome.

Maybe as a Proof of Concept, only focusing on the paradigm shift could make this PR a bit lighter. It feels like:
- src/Playwright/Browser.php
- src/Playwright/BrowserContext.php
- src/Playwright/BrowserType.php
- src/Playwright/Client.php
- src/Playwright/Element.php
- src/Playwright/Frame.php
- src/Playwright/Page.php
- src/Playwright/Playwright.php
- src/Playwright/Server.php

are the kind of the main files responsible for the paradigm shift and we could maybe focus on these first, with maybe one operation/assertion migrated to start with, just to test it, so you don't have to tackle the entire shift on your own.

It's quite a lot to take on and we could all help too once we've worked on the underlying shift. Narrowing the PoC to focus on the underlying shift only could help us to focus on this specific part first before migrating everything across as there is quite a lot going on here.

src/Playwright/Element.php
ijpatricio: 🤣

ijpatricio:
Hey @dzmitry-vasileuski @clemblanco 

I had to loose the fear of touching a mountain of files. We have to move on :) 

The strategy is to now grow from this minimal setup. Check photo below.

Regarding this PR, better to close and port it chunk by chunk, in smaller PR'r, right?

Nothing is written in stone, suggestions are welcome!

nunomaduro:
This is already merged right?

clemblanco:
I believe so yeah.

PR #91, opened 2025-03-24
Fix bold formatting in README

lokomilo:
Bold formatting syntax (**) had spaces causing it to render incorrectly.

### Before
image
image

---

### After
image
image


ijpatricio:
Nice eye, great eye and touch! Thanks!

PR #95, opened 2025-06-07
[WIP] feat: adds new expectation

sonalidudhia:


tests/Pest.php
```diff
@@ -92,4 +92,53 @@ function cleanupScreenshots(): void
 
     return $this;
 });
+
+expect()->extend('toHaveId', function (string $id): Expectation {
+
+    if (! $this->value instanceof Element && ! $this->value instanceof Locator) {
```
MrPunyapal, line 88: No need to add expectations for Element classes now!
I removed it from the existing as well.

PR #99, opened 2025-06-14
feat:expectation toHaveScreenShot add

sonalidudhia:


nunomaduro:
we already have this.

PR #102, opened 2025-06-18
add option to run a browser with ui

nuernbergerA:
`Playwright::withUi()`

first api was `page(withUi: true)` but then it would be true for all instances because we launch 1 browser and inside multiple isolated contexts


laserhybiz:
Why not name it `headed()`?

nunomaduro:
Going todo this: ./vendor/bin/pest --headless

PR #117, opened 2025-08-21
fix: update filename arg to match documentation

jeremyl-dev:
## Description
The named argument `name` for the ```screenshot()``` method does not align with the argument name described in the documentation (`filename`). Updated the named argument to match with the documentation.


nunomaduro:
thanks; fixed already.

PR #118, opened 2025-08-21
Preserve query parameter structure

Frnk1996:
AmpRequest::getQueryParameters() normalizes all values into arrays, which broke query parsing inside Laravel. Updated handleRequest() to build the full path with the raw query string instead, so Laravel receives parameters in their expected scalar/array form.

### DIFF
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -222,22 +222,35 @@ final class LaravelHttpServer implements HttpServer
             Execution::instance()->tick();
         }
 
-        $filepath = public_path($path = $request->getUri()->getPath());
+        $uri = $request->getUri();
+        $path = $uri->getPath() ?: '/';
+        $query = $uri->getQuery() ?? '';
+        $fullPath = $path . ($query !== '' ? '?' . $query : '');
+        $absoluteUrl = rtrim($this->url(), '/') . $fullPath;
 
-        if (file_exists($filepath) && ! is_dir($filepath)) {
+        $filepath = public_path($path);
+        if (file_exists($filepath) && !is_dir($filepath)) {
             return $this->asset($filepath);
         }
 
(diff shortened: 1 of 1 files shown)


mateusjunges:
Ran into this too. Here's a repo for reproducing the error in case it helps: https://github.com/mateusjunges/reproduce-browser-testing-error

src/Drivers/LaravelHttpServer.php
```diff
@@ -222,22 +222,35 @@ private function handleRequest(AmpRequest $request): Response
...
+        $absoluteUrl = rtrim($this->url(), '/') . $fullPath;
+
+        $filepath = public_path($path);
+        if (file_exists($filepath) && !is_dir($filepath)) {
             return $this->asset($filepath);
         }
 
         $kernel = app()->make(HttpKernel::class);
 
+        $contentType = $request->getHeader('content-type')[0] ?? '';
```
mateusjunges: `$request->getHeader()` returns a string or null, no need for the `[0]`.
```suggestion
        $contentType = $request->getHeader('content-type') ?? '';
```

src/Drivers/LaravelHttpServer.php
```diff
@@ -222,22 +222,35 @@ private function handleRequest(AmpRequest $request): Response
...
+            parse_str($rawBody, $parameters);
+        }
+
         $symfonyRequest = Request::create(
-            $path,
-            $request->getMethod(),
-            $request->getQueryParameters(),
+            $absoluteUrl,
+            $method,
+            [$parameters],
```
mateusjunges: No need to wrap `$parameters` in array here as the variable is an array already.
```suggestion
            $parameters,
```

mateusjunges (changes requested):
Left a few suggestions but other than that, this fixes my tests.

nunomaduro:
released.

PR #119, opened 2025-08-21
Feature: add ImageDiffView when the diff from Playwright is empty

ss10sb:
### What:

- [ ] Bug Fix
- [x] New Feature

### Description:

`Page::expectScreenshot`

I noticed that sometimes the Playwright diff comes back empty when the images are very similar (font style differences), yet the internal string comparison will fail since they really are different.  I created a silly question mark image for the diff image and output the diff view file before the non-diff exception.  So now I can see that the numeric one was a slightly different font :)


nunomaduro:
released

PR #120, opened 2025-08-22
Improve detection of explicit selectors

mateusjunges:
## What's Changed

This PR adds the ability for using exact match CSS selectors when clicking, pressing, filling or doing any operations that require a CSS selectors. 

### Problem
I just converted a Dusk test suite to Pest v4 browser testing, and I use [flux UI](https://fluxui.dev) in the application.
One of my pages have a `<flux:switch>` component, which generates a `<ui-switch>` component once the page is rendered. Here's how I select the element for clicking with dusk:

```php
$browser->click('ui-switch[name="monday"]');
```
When doing the same with pest, I get the following exception:

Exception screenshot

### Solution
I've updated the `Selector::isExplicit` method to better detect explicit selectors. It now correctly detects explicit selectors:
- `[name]`: Any element that has the `name` attribute
- `[name*="test"]`: Any element where `name` contains the substring `test`
- `[name^="test"]`: Any element where `name` starts with `test`
- `[name$="test"]`: Any element where `name` ends with `test`
- `div.test`: Any `div` elements that also have the `test` class
- `div#test`: Any `div` elements that also have the `test` id


src/Support/GuessLocator.php
```diff
@@ -45,6 +45,14 @@ public function for(string $selector, ?string $value = null): Locator
             );
         }
 
+        $locator = $this->page->unstrict(
```
nunomaduro: i don't understand; shouldn't this be caught by the `Selector::isExplicit($selector)`?

nunomaduro (changes requested):

src/Support/GuessLocator.php
mateusjunges: @nunomaduro hmm yeah, I think that's a valid point. I improved the `isExplicit` method to handle a few more cases:

- Any string ending in `[]` is not explicit as it's not a valid css selector
- Check for strings containing the `[`, `.`, `#` characters, which cover things like `div#id`, `div.class`, `input[name="test"]`

jasonmccreary:
Came across this as well today. I think this should be "improved" farther. Really if there are any special characters relating to CSS selection, it should be treated as an _explicit selector_.

It's far more likely a developer is typing a selector here, than attempting to click/press something with special characters in the text. Otherwise, I'd expect never-ending PRs "fixing" this until a stance is taken.

mateusjunges:
@jasonmccreary I started with the selectors I'm using on my application and that were easier for me to test, but I 100% agree with you. I just added a new commit that takes a better approach and checks for any CSS selectors. Let met know what you think.

`[`, `]` - attribute selectors
`#` - ID selectors
`.` - class selectors
`>`, `+`, `~` - combinators
`:` - pseudo-classes/pseudo-elements
`*` - universal selector
`|`, `^`, `$`, `=` - attribute selector operators
`,` - selector lists
`(`, `)` - functional pseudo-classes

jasonmccreary:
@mateusjunges, nice. Yeah, in my opinion, only alphanumeric and whitespace strings should be considered "value" matches. Everything else should be treated as a CSS selector. It's a harder line, but much easier to communicate. I'll leave that up to @nunomaduro.

Either way, I hope at least this gets merged as it aligns much more with what a developer expects.

mateusjunges:
@nunomaduro one thing I noticed that will break with this is the ability to `type` in input fields using the input name as the selector. I think we should also update the `type` method to look for input fields only, similar to what https://github.com/pestphp/pest-plugin-browser/pull/134 does for the `press` method.

jasonmccreary:
@mateusjunges, thanks for this PR!

PR #121, opened 2025-08-22
Rename `$name` to `$filename` to match docs

mateusjunges:
The [documentation](https://v4.pestphp.com/docs/browser-testing) says that we can change the name of the screenshot by passing the `filename` parameter to the screenshot method:

CleanShot 2025-08-21 at 11  26 54

However, when I add the `filename` parameter, I get the following error:

CleanShot 2025-08-21 at 11  29 15 2

This PR renames the parameter to match the docs.


ju5t:
As a reference: this is also PR'd in #117.

PR #124, opened 2025-08-23
add ability to change userAgent per test

jackbayliss:
Would be nice to change user agent per test if needed, since playwright gives us the ability.  Same as withLocale, withTimezone etc.

Will do another PR to set it as an overall option similar to headed maybe 🤔 as would be nice.

See playwright setting: https://playwright.dev/docs/api/class-testoptions#test-options-user-agent


cyppe:
I came here for this, hope this gets merged. Much needed.

jasonmccreary:
@jackbayliss, thanks for this PR!

PR #126, opened 2025-08-23
userAgent configuration setting

jackbayliss:
Adds the ability to use `pest()->browser()->userAgent('...');` which then sets the userAgent for all tests.  Same as headless etc.

Thought it would be cool to have the ability to set it overall.


jasonmccreary:
@jackbayliss, thanks for this PR!

PR #127, opened 2025-08-23
Adding support for Geolocation emulation

nexxai:
## Problem

I have a form that requires GPS coordinates be fetched and included, using JavaScript's getCurrentPosition API.

## Proposed solution

As described in the [Playwright docs](https://playwright.dev/docs/emulation#geolocation), this PR allows you to chain a `->geolocation(1.2345, 6.78910)` call in your visit which will set the testing browser's location to those coordinates.

Example:
```php
visit('/')
    ->geolocation(51.5076, 1.1124)
    ->assertSee(...)
```

## Notes

If this PR is accepted, https://github.com/pestphp/docs/pull/322 will update the docs to reflect these proposed changes

### DIFF
--- a/src/Api/PendingAwaitablePage.php
+++ b/src/Api/PendingAwaitablePage.php
@@ -107,6 +107,18 @@ final class PendingAwaitablePage
     }
 
     /**
+     * Sets the geolocation for the page.
+     */
+    public function geolocation(float $latitude, float $longitude): self
+    {
+        return new self($this->browserType, $this->device, $this->url, [
+            'geolocation' => ['latitude' => $latitude, 'longitude' => $longitude],
+            'permissions' => ['geolocation'],
+            ...$this->options,
+        ]);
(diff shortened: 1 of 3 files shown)


nunomaduro:
any chance we can do something like: visit()->from()->newYork() or osmething lkie that? like the `on` method? so we have most popular cities?

nunomaduro:
can you make a pull request that adds this? maybe add the other methods first, like locale, timezone, etc. but after add the cities if you can: https://x.com/enunomaduro/status/1959662770858926404.

nexxai:
Hey @nunomaduro just wondering if instead of '->from()->newYork()' we should do '->fromNewYork()' as a magic method?

If not, where should I add the individual city methods?  I've never built something like that and I assume they should live in a separate cities class but I'm not sure how we would call it from the 'PendingAwaitablePage'.

PR #128, opened 2025-08-24
feat: allow overriding the request hostname

ju5t:
We run a multitenant application where we must specify a hostname, without it, the app doesn't know what to do and throws a 500 error. This PR makes it possible to add the hostname to the request that gets sent in the http server.

```php
pest()->browser()->withHostname('test.localhost');
```

I'm not 100% if this is the right approach and I'm sure some things are missing (e.g. rewrites, assertions maybe?) but I wanted to at least get this on the radar as I know others will need this too.

If anything doesn't fit the current standards or thoughts behind this package let me know.

And thanks for such an amazing tool, looking forward to using it :)


ju5t:
This PR is ready for review. Let me know if anything needs to be changed.

The `APP_KEY` in phpunit.xml is required, without it the `$kernel->terminate()` call inside the http server would throw an exception that the encryption key was the wrong length.

jasonmccreary:
@ju5t, thanks for this PR!

ju5t:
@jasonmccreary cool! Do you still need anything from me to get this merged?

jasonmccreary:
@ju5t, nope. Was just waiting for the tests to pass.

ju5t:
@nunomaduro I see you reverted this PR in your recent commits. Could you give some feedback on what could be improved? 

We can't use browser testing unless we can specify a hostname to test on. We've written tests on the assumption this would not be reverted (or if it would, it would be improved). From what I can see at the moment, there is no alternative for it.

I'm also OK if you don't see any use for this, but in that case I think it would've been better to not merge the PR.

If you can let me know what your thoughts are on this, I'm happy to send an improved PR!

pindab0ter:
For what it's worth, this PR would not have solved my issue: https://github.com/pestphp/pest/issues/1520

PR #130, opened 2025-08-25
Adding `from()->{cityName}()` chainable method to automatically set geolocation, locale, and timezone

nexxai:
## Problem

@nunomaduro [tweeted](https://x.com/enunomaduro/status/1959662770858926404) that it would be awesome that instead of having to pass in hardcoded lat/long coordinates, a person could just provide a popular city and it would automatically set their geolocation, locale, and timezone.  I agree.  This would be sick.

## Solution

I've built a `from()` method that accepts a chained method of the names of several popular cities worldwide, as listed in `src/Enums/Cities.php`.  

The value of the enum (e.g. "newYork") is accepted as the name of the method and then uses the existing `->geolocation()`, `->withLocale()`, and `->withTimezone()` methods to set them accordingly.  All one needs to do add a city is to update the enum and the relevant match statements below.

For example, you could do:

```php
visit('/')
    ->from()->newYork()
    ...
```

or 

```php
visit('/')
    ->from()->amsterdam()
    ...
```

etc.


c0nst4ntin:
Nice pull request (and fast). I was just about to create a similar pull request, when I saw this. 

I am unsure if I wouldn't prefere `->from(Cities::AMSTERDAM);` to `->from()->amsterdam()` since the parameter could be typed it might result better ide support. Maybe it would also make it more extendable, incase someone wants to add a new city, they could perhaps overwrite the Enum.

But since this is exactly how @nunomaduro suggested it, we will see what he says.

nexxai:
After discussions with @jasonmccreary I have added the ability to also pass a city from the `Cities` enum directly to the `->geolocation()` method for those who prefer to skip the magic and get the benefits of better IDE support.

Example:
```php
use Pest\Browser\Enums\Cities;

visit('/')
    ->geolocation(Cities::NEW_YORK)
    ...
```

jasonmccreary:
@nexxai, I think overloading the `geolocation` method is a good balance. Maybe just see if there is a way to refactor the code so the signature is a bit tighter.

nexxai:
Looking into options right now.  Thanks for this feedback

nexxai:
I went and reviewed common overloading practices and I believe this is best option in terms of tightening up the signature.  I renamed the variable to be less clunky to read and also added some usage examples.

I'm definitely open to hearing any other suggestions if anyone has any

jasonmccreary:
I think this looks good for what the tweet outlines and balances the PR comments to allow more use of the `Cities` enum. While there may be many, future PRs would make small changes to add more `Cities` in isolation.

PR #133, opened 2025-08-27
Correct value() docblock and add getPlaywrightPage() accessor

Frnk1996:
Corrected the value() docblock to reflect the method's purpose.

Added a getPlaywrightPage() method to expose the underlying Playwright Page instance, since the docblock on the value() method suggests it either used to be there or was supposed to be there.

The method is intentionally called getPlaywrightPage() instead of page() to avoid confusion with Pest’s own WebPage instance (avoids $page->page() in tests).

### DIFF
--- a/src/Api/Webpage.php
+++ b/src/Api/Webpage.php
@@ -96,7 +96,15 @@ final readonly class Webpage
     }
 
     /**
-     * Gets the page instance.
+     * Gets the Playwright page instance.
+     */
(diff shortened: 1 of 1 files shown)


Frnk1996:
Just found the page() method on AvailableWebPage, no clue how I missed that yesterday. Don't know if the getPlaywrightPage() method makes sense now.

PR #141, opened 2025-09-03
Correct undefined SIGTERM error when running in a Linux environment.

zfi:
PROBLEM

When running tests on the project from a Gitlab CI runner, the tests all succeed but the test suite fails at the very end with an undefined constant 'SIGTERM' error.

```php
  Undefined constant "Pest\Browser\Playwright\Servers\SIGTERM"
  at vendor/pestphp/pest-plugin-browser/src/Playwright/Servers/PlaywrightNpmServer.php:99
     95▕     {
     96▕         if ($this->systemProcess instanceof SystemProcess && $this->isRunning()) {
     97▕             $this->systemProcess->stop(
     98▕                 timeout: 0.1,
  ➜  99▕                 signal: PHP_OS_FAMILY === 'Windows' ? null : SIGTERM,
    100▕             );
    101▕         }
    102▕
    103▕         $this->systemProcess = null;
  ```

This is running version 4.0.3 of the pest-plugin-browser. The CI runner is loading a Ubuntu 24.04 Docker container, running PHP 8.4. 

The constant 'SIGTERM' is declared in the ext-pcntl package, which is currently loaded as a dev dependancy. When the plugin is released, the pcntl package is no longer included, and the constant definition is lost.

SOLUTION

Moving the pcntl requirement from require-dev to require will include the package in the released plugin.


jasonmccreary:
Hey, thanks for your PR. I'm helping get caught up on some of these PRs. However, I am going to leave this up to @nunomaduro.

biinari:
Alternatively, the signal could be omitted to use the default. [Symfony\Component\Process\Process->stop()](https://github.com/symfony/process/blob/7.3/Process.php#L908) uses 15 (`SIGTERM`) followed by 9 (`SIGKILL`) if it has not stopped before the given timeout. This would be slightly better as it can be stopped properly if the playwright process fails to stop on the initial `SIGTERM` signal.

Along with this, it might make sense to increase the timeout to 1 second to give playwright opportunity to stop cleanly.

The comment in Symfony Process indicates the choice to use integer values directly in case the constants are not defined.

PR #145, opened 2025-09-04
feat: add support for taking screenshot of an element

mansoorkhan96:
This pull requests solves https://github.com/pestphp/pest/issues/1481 by adding the support to take screenshot of an element.

Example:

```php
$page = visit('/');

$page->screenshotElement('#screenshot-element', 'element-screenshot.png');
```

### DIFF
--- a/src/Api/Concerns/CanTakeScreenshots.php
+++ b/src/Api/Concerns/CanTakeScreenshots.php
@@ -1,0 +1,37 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Api\Concerns;
+
+trait CanTakeScreenshots
+{
+    /**
+     * Performs a screenshot of the current page and saves it to the given path.
+     */
+    public function screenshot(bool $fullPage = true, ?string $filename = null): self
+    {
+        $filename = $this->getFilename($filename);
+
+        $this->page->screenshot($fullPage, $filename);
+
+        return $this;
+    }
+
+    /**
+     * Performs a screenshot of an element and saves it to the given path.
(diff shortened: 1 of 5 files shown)


jasonmccreary:
@mansoorkhan96, can you resolve the conflicts here and we'll take another look.

mansoorkhan96:
@jasonmccreary resolved!

jasonmccreary:
@mansoorkhan96, for [Symmetry](https://laracasts.com/episodes/3644), what do you think about calling this `InteractsWithScreen` rather than `CanMakeScreenshots`.

mansoorkhan96:
> @mansoorkhan96, for [Symmetry](https://laracasts.com/episodes/3644), what do you think about calling this `InteractsWithScreen` rather than `CanMakeScreenshots`.

I like `InteractsWithScreen` and how i did not know about this course. Thanks man

jasonmccreary:
@mansoorkhan96, this test appears flaky. My guess is since it deals with dates, you may need to freeze time.

mansoorkhan96:
> @mansoorkhan96, this test appears flaky. My guess is since it deals with dates, you may need to freeze time.

I guess thats only possible with Carbon?

jasonmccreary:
Yes. I assumed it was wrapped in Pest. Otherwise, you may want to reconsider your suffix. Why does it need a date? Shouldn't another run overwrite with the filename they provided?

jasonmccreary:
@mansoorkhan96, nice. Just make sure to note in the docs that the default filename is the test name.

PR #146, opened 2025-09-04
feat: add support for passing params to click

mansoorkhan96:
This pull requests solves https://github.com/pestphp/pest/issues/1477 by adding support for passing params to `click`.

Example:

```php
$page = visit('/');

$page->click('#button');
$page->assertDontSeeIn('#result', 'Option-2 clicked');

// clickCount => 2 is considered as double click
$page->click('#button', options: ['clickCount' => 2]);
$page->assertSeeIn('#result', 'Option-2 clicked');
```

### DIFF
--- a/src/Api/Concerns/InteractsWithElements.php
+++ b/src/Api/Concerns/InteractsWithElements.php
@@ -14,9 +14,9 @@ trait InteractsWithElements
     /**
      * Click the link with the given text.
      *
-     * @param  array<string, mixed>|null  $options
+     * @param  array<string, mixed>  $options
      */
-    public function click(string $text, ?array $options = null): Webpage
+    public function click(string $text, array $options = []): Webpage
     {
         $this->guessLocator($text)->click($options);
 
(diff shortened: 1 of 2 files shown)


mansoorkhan96:
I believe the check randomly failed on something weird.

jasonmccreary:
@mansoorkhan96, I've merged a bunch of PRs. Can you resolve the conflicts and ensure all tests are passing.

jasonmccreary:
@mansoorkhan96, thanks.

I'm not sure what `click` accepts underneath, but wouldn't a default of an empty array be more appropriate for `$options` instead of `null`?

mansoorkhan96:
> @mansoorkhan96, thanks.
> 
> I'm not sure what `click` accepts underneath, but wouldn't a default of an empty array be more appropriate for `$options` instead of `null`?

I wanted to add typed array in doc block for options by looking at Playwright docs but then i realised we might not catch up with their changes. I can send a PR that adds proper typed array for all the options params in project, if its the way to go?

I will set empty array as default for options that would be correct.

jasonmccreary:
@mansoorkhan96, I don't think that's necessary at this point.

I just wanted to isolate this single parameter type since it is new and it seems to be an empty array is a more reasonable default than `null`.

PR #150, opened 2025-09-05
Fix string formatting in console assertion messages

emargareten:
When a url contains reserved characters it gets percent-encoded which breaks the expected number of arguments for sprintf and causes an ArgumentCountError. This PR fixes this issue by moving the url into a separate argument instead of including it directly in the format string.


jasonmccreary:
@emargareten, nice catch. Thanks!

PR #153, opened 2025-09-05
tests: add an additional test waitForSelector

mansoorkhan96:
This PR adds an additional test for `waitForSelector` to make sure it can accept options.

### DIFF
--- a/tests/Browser/Webpage/WaitTest.php
+++ b/tests/Browser/Webpage/WaitTest.php
@@ -56,3 +56,26 @@ it('may wait for selector', function (): void {
 
     $page->assertSeeIn('#child', 'Child Element Added');
 });
+
+it('may wait for selector with options', function (): void {
+    Route::get('/', fn (): string => '
+        <div id="container">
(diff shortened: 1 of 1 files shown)


jasonmccreary:
@mansoorkhan96, are there multiple states for this? If so, might be nice to call out each in a test.

jasonmccreary:
@mansoorkhan96, maybe I don't understand the _state_, but it seems odd that of them are the same test. Wouldn't _visible_ or _attached_ be a different state when an element is removed?

Feel free to link me to the Playwright docs so I can familiarize myself with the intent.

mansoorkhan96:
> @mansoorkhan96, maybe I don't understand the _state_, but it seems odd that of them are the same test. Wouldn't _visible_ or _attached_ be a different state when an element is removed?
> 
> Feel free to link me to the Playwright docs so I can familiarize myself with the intent.

My bad, I have added better tests now.

jasonmccreary:
@mansoorkhan96, thank you for being thorough. These are great.

PR #155, opened 2025-09-05
Adds `within` method to the webpage

mateusjunges:
This PR adds a new `within()` method to the Webpage class that allows scoping browser interactions to specific CSS selectors.

```php
$page->within('#contact', function ($browser) {
    $browser->click('Submit');
});
```

## Usage

The `within()` method accepts a CSS selector and a callback function, scoping all browser interactions within that callback to the specified container:

```php
// Click a button within a specific form
$page->within('#contact-form', function ($page) {
    $page->click('Submit Button');
});

// Type in a form within a modal dialog
$page->within('.modal', function ($page) {
    $page->type('email', 'user@example.com')
         ->type('password', 'secret')
         ->press('Login');
});

// Assert text exists within a specific section
$page->within('#header', function ($page) {
    $page->assertSee('Welcome');
    $page->assertDontSee('Footer Content');
});
```

### Some tech details
I've decided to remove the `readonly` from `Webpage`.  This is requried to store the current scope as an instance property. If you check the commit history, I considered adding a new class for handling state, but I think it adds unecessary complexity. Also, I think the "scope" state actually belongs to the webpage itself. Open to suggestions if y'all think this implementation is not good 🙂


mateusjunges:
Been tinkering with a `ScopedWebpage` instead of using the CSS `>>` operator, but curious to hear feedback on this before I spend more time on it.

jasonmccreary:
@mateusjunges, love this! Is it ready to merge? Your last comment made me think you have more you want to do... If so, please mark it as draft.

src/Api/Concerns/MakesElementAssertions.php
```diff
@@ -42,13 +42,11 @@ public function assertTitleContains(string|int|float $title): Webpage
...
      */
-    public function assertSee(string|int|float $text): Webpage
+    public function assertSee(string|int|float $text): self
     {
         $text = (string) $text;
 
-        $locator = $this->page->unstrict(
-            fn () => $this->page->getByText($text),
-        );
+        $locator = $this->getTextLocator($text);
```
mateusjunges, line 49: This was extracted and is now a method on `Webpage` and `ScopedWebpage`, and it's required for scoping the selector when using `within`. The `guessLocator` already allows for this in other assertions/interactions, but this selector was hardcoded in here.

mateusjunges:
@jasonmccreary while reading more about the `>>` CSS operator, I discovered it was deprecated and later on removed from modern CSS specifications. This lead me to explore a better implementation, but then I noticed that this operator is already used in other places in the codebase:

https://github.com/pestphp/pest-plugin-browser/blob/e2557a8093cc78e5d5ff50f7845377d2d74a1dc9/src/Playwright/Locator.php#L404

I'll leave it up to you and Nuno to decide whether it's an issue or not. The other implementation I had involves a `ScopedWebpage`, which would scope the selector using the Locator class (which uses `>>` under the hood, so I guess it's the same thing). I also would have to change the return type on pretty much all traits from `Webpage` to `self`, as they would be reused on the `ScopedWebpage`.

Anyway, I think this is ready for review 😅

mansoorkhan96:
Great work @mateusjunges

jasonmccreary:
@mateusjunges, if it's just in other places, then I feel it's fine to continue to use for now. Playwright doesn't have a direct instance of `within`. From some searching, it seems most implementations just prepend the original `within` selector. From a CSS standpoint it would work (descendent selector).

For example:

```diff
- return new self($this->frameGuid, $this->selector.' >> '.$selector);
+ return new self($this->frameGuid, $this->selector.' '.$selector);
```

Again, I don't think this change needs to be made now. It can be made as a future refactor to remove the `>>` everywhere.

PR #163, opened 2025-09-10
feat: add ability to interact with dialogs

mansoorkhan96:
This pull request adds the ability to test and interact with browser dialogs. Playwright automatically dismisses the dialogs if they are not explicitly interrupted in tests. However, sometimes you might want to make sure if certain dialogs appear, under certain conditions and maybe with some dynamic content or alert.

Example: 

```php
use Pest\Browser\Playwright\Dialog;

$page = visit('/');

$page->onDialog(function (Dialog $dialog): void {
    expect($dialog->type())->toBe('alert');
    expect($dialog->message())->toBe('Hello World!');

    $dialog->accept();
});

$page->click('#alert-btn');
```

### DIFF
--- a/src/Api/Concerns/InteractsWithDialogs.php
+++ b/src/Api/Concerns/InteractsWithDialogs.php
@@ -40,7 +40,7 @@ trait InteractsWithDialogs
     /**
      * Set up automatic dialog acceptance for all future dialogs.
      */
-    public function acceptingDialogs(?string $promptText = null): self
+    public function acceptAllDialogs(?string $promptText = null): self
     {
         $this->page->onDialog(function (Dialog $dialog) use ($promptText) {
             $dialog->accept($promptText);
@@ -52,7 +52,7 @@ trait InteractsWithDialogs
     /**
      * Set up automatic dialog dismissal for all future dialogs.
      */
-    public function dismissingDialogs(): self
+    public function dismissAllDialogs(): self
     {
         $this->page->onDialog(function (Dialog $dialog) {
             $dialog->dismiss();
--- a/src/Api/Concerns/InteractsWithDialogs.php
+++ b/src/Api/Concerns/InteractsWithDialogs.php
@@ -1,0 +1,95 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Api\Concerns;
+
+use Closure;
+use Pest\Browser\Playwright\Dialog;
+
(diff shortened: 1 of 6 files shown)


mansoorkhan96:
@jasonmccreary Would love to get your feedback on this, if you have time.

biinari:
This is really useful in testing an app that has many confirm dialogs. There's just a little thing that has been tripping me up.

When an expectation in the dialog handler function fails, it can cause a hang where the dialog remains open (with no information on which test case is running). This happens if the interaction that opened the dialog uses an automatic selector.

Demonstration with a test case:

```php
it('does not hang when dialog expectation fails', function (): void {
    Route::get('/', fn (): string => '
        <button id="confirm-btn" onclick="confirm(\'Are you sure?\');">Show Confirm</button>
    ');

    $page = visit('/');

    $page->onDialog(function (Dialog $dialog): void {
        expect($dialog->message())->toBe('Wrong text');

        $dialog->accept();
    });

    // $page->click('#confirm-btn'); // times out as desired and fails with "Timeout 1000ms exceeded"
    $page->click('Show Confirm'); // makes test hang with dialog still open
})->throws(ExpectationFailedException::class);
```

This can be worked around by a try / catch in the dialog handler:

```php
    $page->onDialog(function (Dialog $dialog): void {
        try {
            expect($dialog->message())->toBe('Wrong text');
        } catch (ExpectationFailedException $e) {
            $dialog->dismiss();
            throw $e;
        }

        $dialog->accept();
    });
```

That also gives a nicer result of making the test fail with the original expectation failure rather than a timeout for the click, i.e. "Failed asserting that two strings are identical." (with a diff).

Does this try / catch inside the handler feel too long-winded for inclusion in example usage?

Would it be appropriate to catch this exception in `Pest\Browser\Playwright\Page->handleDialogEvent()`, dismiss the dialog if it has not yet been dismissed / accepted, and re-throw the exception?

mansoorkhan96:
@biinari I will look into this.

mansoorkhan96:
@biinari Added the try catch to properly avoid test freezing issue.
Thanks for the feedback.

EdgarsK:
Any updates on when this will be merged?

phenix-factory:
I do need to feature to remove a hack that bypass the dialog in my project. I would love to see this moving forward a bit :)

aon4o:
+1

PR #164, opened 2025-09-12
Allow to limit interactions `within` a scope

mateusjunges:
This PR is another stab at #155, and it adds a new `within()` method to the Webpage class that allows scoping browser interactions to specific CSS selectors.

```php
$page->within('#contact', function ($browser) {
    $browser->click('Submit');
});
```

### More details
I see that the first PR got reverted because it made the `Webpage` non-readonly. I copied the tests from my previous PR and reworked the implementation. 

I changed the return type on the `InteractWithElements`, `MakesElementAssertions` and `MakesConsoleAssertions` from `Webpage` to `self`, as those traits are now also used on the `ScopedWepage` class.

The new `ScopedWebpage` now keeps context of the scope, so no changes to the `Webpage` are required.

### DIFF
--- a/src/Api/Concerns/InteractsWithElements.php
+++ b/src/Api/Concerns/InteractsWithElements.php
@@ -100,7 +100,7 @@ trait InteractsWithElements
     /**
      * Right-click the element matching the given selector.
      */
-    public function rightClick(string $text): Webpage
+    public function rightClick(string $text): self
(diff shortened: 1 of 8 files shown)


mateusjunges:
Sad to see this sitting idle for 2 months without any review 🙁

PR #165, opened 2025-09-17
[Bug] `Selector::isExplicit()` assume any `.` refers to CSS selector

crynobone:
This only illustrates the issue and a proper fix needs to be implemented:

image

PR #167, opened 2025-09-19
Fix: class selector

MrPunyapal:
Fixes: https://github.com/pestphp/pest-plugin-browser/pull/165


crynobone:
This should do it, but I do also wonder if there's any other CSS selector that should be ignored when followed with space.

PR #169, opened 2025-09-26
Fix stopping playwright server

biinari:
When pest completes, a terminate signal is sent to stop the playwright server. As it is run using a shell (by passing commandline as a string), the shell receives this signal, which it does not forward to playwright.

This means that the `playwright run-server` command continues running after pest exits.

## Solution

Run playwright server command without shell (passing commandline as an array) so that signals can be sent directly to playwright to terminate it.

## Related change

Avoid persisting a new port for `AlreadyStartedPlaywrightServer` if playwright is already set.

### DIFF
--- a/src/Playwright/Servers/PlaywrightNpmServer.php
+++ b/src/Playwright/Servers/PlaywrightNpmServer.php
@@ -60,11 +60,11 @@ final class PlaywrightNpmServer implements PlaywrightServer
             return;
         }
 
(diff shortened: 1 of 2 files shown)


cyppe:
This is important, as we found 300 orphaned playwright/chromium inside ddev after some time.

Had to override it hard to fix it until this is merged and released,.

image

PR #172, opened 2025-10-01
Remove need for pcntl extension

biinari:
Alternative to #141 to fix running on a system that does not have `pcntl` extension enabled.

## Solution

Stop passing `SIGTERM` (defined in `pcntl` extension) to `Symfony\Component\Process\Process->stop()`, allowing it to use the default. This default behavior is to signal with `SIGTERM` and if the process fails to stop before the timeout, signal again with `SIGKILL`.

I considered increasing the timeout to ensure playwright server has sufficient time to stop from the initial `SIGTERM` signal. My measurements show it stopping in under 1 millisecond so 0.1 seconds seems ample.


stancl:
Would appreciate if this could get merged.

We have an issue in one of our apps where browser tests generally work fine on their own but when executed in a specific way in parallel, some testing processes just hang and never finish.

As mentioned in this PR, passing `SIGTERM` is incorrect here because what the `stop()` method does is that it *tries* to stop the process gracefully with SIGTERM first and if that doesn't work, it retries with a SIGKILL. Sending consecutive SIGTERMs doesn't appear to do anything about a hanging process here.

biinari:
I'm happy to rebase or merge to resolve the conflicts in `composer.json` (due to neighbouring lines changed) if that helps to get this change merged in.

PR #178, opened 2025-10-12
feat: Assert no console messages at given level(s)

biinari:
Add `assertNoConsoleMessages(array|string $levels)`

Check page for console messages at specified levels:

`'debug'`: `console.debug()` calls (not checked by default)
`'error'`: `console.error() `calls
`'info'`: `console.info()` calls
`'warning'`: `console.warn()` calls (can also specify level as `'warn'`)

Usage:

```php
// assert no console messages have been made at any level
$page->assertNoConsoleMessages(['debug', 'error', 'info', 'warning']);

// assert no console.warn() calls have been made
$page->assertNoConsoleMessages('warning');

// get list of console.warn() and console.error() calls made on page
$page->consoleMessages(['error', 'warning']);
```

Note: I have not added this assertion to `assertNoSmoke()` as it may not be desired by every project.

### DIFF
--- a/src/Api/Concerns/MakesConsoleAssertions.php
+++ b/src/Api/Concerns/MakesConsoleAssertions.php
@@ -75,7 +75,7 @@ trait MakesConsoleAssertions
     /**
      * Asserts there are no console messages at given levels on the page.
      *
-     * @param array<int, string>|string $levels Console message level(s) to check for (debug, info, error, warning).
+     * @param  array<int, string>|string  $levels  Console message level(s) to check for (debug, info, error, warning).
      */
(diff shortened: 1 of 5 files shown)


Gregory-Gerard:
Friendly +1 on this one 👋

I ran into this recently, I thought `assertNoJavaScriptErrors()` would also catch `console.error` calls, but it doesn't. In my app a lot of real issues go through `console.error` (missing react-i18n translation keys, Apollo `onError`, React warnings...) and they were all silently ignored in my tests.

For now I'm patching `console.error` in a custom init script to push into `window.__pestBrowser.jsErrors`, but having `assertNoConsoleMessages('error')` natively would be way better. Would be great to see this merged.

PR #184, opened 2025-11-03
Click method option add

### LINKED ISSUE #1518: Documentation for the $page->click() method say it accepts options.
Unless I am missing something here, the docs say this for the click method:

```
You may also pass options:

$page->click('#button', options: ['clickCount' => 2]);
```

But the method signature for click is the following:

```
 /**
     * Click the link with the given text.
     */
    public function click(string $text): self
    {
        $this->guessLocator($text)->click();

        return $this;
    }
```

Are the docs or the method signature wrong? From everything I can tell, the click method doesn't accept an options argument.

sonalidudhia:
Resolves https://github.com/pestphp/pest/issues/1518

### DIFF
--- a/src/Api/Concerns/InteractsWithElements.php
+++ b/src/Api/Concerns/InteractsWithElements.php
@@ -130,7 +130,7 @@ trait InteractsWithElements
 
         $currentValue = $locator->inputValue();
 
-        $locator->fill($currentValue . $value);
+        $locator->fill($currentValue.$value);
 
         return $this;
     }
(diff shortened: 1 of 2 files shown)


sonalidudhia:
Above tests failed is not related to my code.

nunomaduro:
sorry, not really convinced this $options array should even exist.

riddla:
@nunomaduro, as the [Playwright click method does have options](https://playwright.dev/docs/input#mouse-click), can you elaborate?

PR #189, opened 2025-12-30
[4.x] Add missing route facade import to test

jackbayliss:
Noticed this while poking about, just helps for IDE completion and what not- used elsewhere so keeps it consistent.


jackbayliss:
CI failing is unrelated to this, but have addressed it via https://github.com/pestphp/pest-plugin-browser/pull/190

PR #195, opened 2026-01-11
Add optional screenshot scale support

WINBIGFOX:
Add optional screenshot scale support across the page and element APIs, defaulting to css when scale is omitted or null, wire the value into Playwright screenshot options, and add browser tests covering device scale screenshots to keep existing behavior intact.

### DIFF
--- a/src/Api/Concerns/InteractsWithScreen.php
+++ b/src/Api/Concerns/InteractsWithScreen.php
@@ -9,11 +9,11 @@ trait InteractsWithScreen
     /**
      * Performs a screenshot of the current page and saves it to the given path.
      */
-    public function screenshot(bool $fullPage = true, ?string $filename = null): self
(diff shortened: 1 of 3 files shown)


mishahawthorn:
+1 on this, I'm currently patching locally to be able to get device scale for screenshots and was about to PR myself.

The majority of users will want `css` for consistent diffing across platforms, but there are valid use cases to want `device` scaling. In controlled CI environments I can see a desire for "retina-level" assertions and diffing for apps that call for it.

In my use case, I actually maintain a separate test suite that I use to generate screenshots for marketing and help article content.

@nunomaduro Thoughts on expanding the API in this way? One concern I see as a maintainer is the extent of future customization that will be called for and might warrant thinking about the API a bit differently compared to just adding another param.

PR #198, opened 2026-02-04
Add playwright tracing support

Jonathan-384481:
This pull request adds playwright tracing support with a low level API closely following playwright interface. A higher level API accessible from webpage and command line options similar to playwright.

High level API:
```
$page = visit('/');
$page->startTracing();
// Do stuff
$page->stopTracing();
```
or it can be started with command line options
`pest --trace on`
`pest --trace retain-on-failure`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ coverage.xml
 # Playwright
 node_modules/
 /tests/Browser/Screenshots
+/tests/Browser/Tracing
 
 # MacOS
 .DS_Store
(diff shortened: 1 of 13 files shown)


aguingand:
A must-needed feature, especially for debugging in CI. I'm wandering if `Traces` would be better than `Tracing` directory name.

gyulavoros:
`pest --trace retain-on-failure` would be essential for us as well.

PR #199, opened 2026-02-06
fix: prevent JS asset truncation for files larger than 8KB

riasvdv:
## Summary

- Fixes JS asset serving in `LaravelHttpServer::asset()` where files >8KB could be silently truncated
- The old code used `fread()` + `ReadableResourceStream` which has two bugs: `fread()` returns at most 8,192 bytes per call inside Amp's event loop, and `ReadableResourceStream` causes the `content-length` header to be stripped
- The fix uses `file_get_contents()` and returns the content as a string body directly, which correctly sets `content-length` and avoids truncation

## The Bug

The `asset()` method served JS files through this pipeline:

```php
$temporaryContent = fread($file, (int) filesize($filepath));
// ... write to php://temp ...
return new Response(200, [...], new ReadableResourceStream($file));
```

**Three issues:**
1. **`fread()` short-read** — Inside Amp's non-blocking event loop, `fread()` returns at most 8,192 bytes per call, not the full file
2. **Missing `content-length`** — `Response::setBody(ReadableStream)` calls `removeHeader('content-length')`, forcing chunked transfer encoding
3. **`stream_socket_shutdown()` on `php://temp`** — `ReadableResourceStream::close()` calls `stream_socket_shutdown()` on non-socket streams (undefined behavior)

This caused Craft CMS Control Panel pages to fail loading because their JS bundles (~50KB+) were truncated to 8KB.

## The Fix

```php
fclose($file);
$rawContent = file_get_contents($filepath);
$content = $this->rewriteAssetUrl($rawContent);
$response = new Response(200, [], $content);
$response->setHeader('Content-Type', $contentType);
return $response;
```

- `file_get_contents()` reads the entire file regardless of event loop context
- String body → `ReadableBuffer` → correct `content-length` header
- No stream cleanup issues

## Tests

Three regression tests added:
- **`serves JS assets as string body instead of stream`** — Verifies response body is `ReadableBuffer` (not `ReadableResourceStream`) via reflection. Fails with old code
- **`sets content-length header on JS asset responses`** — Verifies `content-length` header matches file size. Fails with old code
- **`serves large JS files completely`** — End-to-end test via Playwright: creates >8KB JS file, loads it in a page, verifies a marker variable at the end of the file is accessible

### DIFF
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -99,12 +99,12 @@ final class LaravelHttpServer implements HttpServer
             return;
         }
 
-        $this->socket = $server = SocketHttpServer::createForDirectAccess(new NullLogger());
+        $this->socket = $server = SocketHttpServer::createForDirectAccess(new NullLogger);
 
         $server->expose("{$this->host}:{$this->port}");
         $server->start(
             new ClosureRequestHandler($this->handleRequest(...)),
-            new DefaultErrorHandler(),
+            new DefaultErrorHandler,
         );
     }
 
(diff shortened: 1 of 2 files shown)


jjjrmy:
having this same issue

riasvdv:
See #208

PR #206, opened 2026-04-06
feat: add video on test failure

veljkho:
Adds support for video recording of browser test sessions, automatically saving a `.webm` video when a test fails. Videos from passing tests are silently deleted to avoid unnecessary disk usage. This is optional just like in Playwright.

Playwright has a native `recordVideo` context option that writes a `.webm` file to a directory during the session. The video is finalized when the browser context closes — which already happens in the existing `afterEach` hook via `Playwright::reset()`. After reset, the plugin moves the video to `tests/Browser/Videos/` on failure, or removes it on success.

Usage

Enable globally in `Pest.php`:

```
browser()->recordVideoOnFailure();
```

or call it via `--record-video` when running tests.

Videos are .webm format, playable in all modern browsers
No new dependencies — relies entirely on Playwright's built-in recording
Parallel runs are safe since each worker uses its own isolated temp directory

Added unit tests.
Tested locally with a local browser test that intentionally failed.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ coverage.xml
 # Playwright
 node_modules/
 /tests/Browser/Screenshots
+/tests/Browser/Videos
 
 # MacOS
 .DS_Store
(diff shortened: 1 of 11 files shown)


nexxai:
Holy crap, yes!  This would be amazing.

@nunomaduro please merge!

PR #208, opened 2026-04-09
fix: use file_get_contents to prevent large JS files (>8KB) from being truncated

### LINKED ISSUE #1635: [Bug]: VITE 8 - fread on 0 bytes files causes 500 error during browser testing
### What Happened

After i build my assets all the browser tests fail with this error.

FAILED  Tests\Tenant\Customers\CustomerBrowserTest > it can edit a customer through the browser
  Timeout 5000ms exceeded. A screenshot of the page has been saved to [Tests/Browser/Screenshots/it_can_edit_a_customer_through_the_browser].

The generated screenshot is a white page. After digging deeper with the help of claude it found an issue on some js routes. Some files of 0 bytes here the reported explanation: 
>   The error only happens in Pest Browser because it uses an in-process Amp HTTP server (LaravelHttpServer) that has a bug: it calls fread($file, 0) on JS files to rewrite
  asset URLs. PHP 8.4 throws a ValueError on fread with length 0, so instead of returning an empty 200, it crashes with a 500.

After its patch in `src/Drivers/LaravelHttpServer.php` i applied temporarly in my vendor files that check the size of the file before the fread and tests are passing correctly.

### How to Reproduce

Any browser tests that checks rendered javascript using vite 8 prod build. Not sure it happens also with vite 7. I also use shadcn vue and claude is saying the some index.ts that just re-exports components becomes empty files during the build.

### Sample Repository

_No response_

### Pest Version

4.3.2

### PHP Version

8.4

### Operation System

macOS

### Notes

_No response_

### LINKED ISSUE #1664: [Bug]: Large JS files (>8KB) are truncated to 8192 bytes in browser tests causing SyntaxError
### What Happened

When running browser tests, all large JavaScript files (>8KB) are truncated to exactly **8192 bytes** by the in-process AMPHP test server, causing `SyntaxError: Unexpected end of input` and `SyntaxError: Invalid or unexpected token` in the browser.

The errors appear on all pages that load large JS assets (Livewire, Filament, Vite bundles). Pages where all JS happens to be <8KB pass silently.

### Evidence

Capturing `window.__pestBrowser.jsErrors` (with filename/line/col) after `waitForLoadState('networkidle')` shows:

```json
[
  { "message": "Uncaught SyntaxError: Unexpected end of input",  "filename": ".../vendor/livewire/livewire.min.js", "lineno": 1, "colno": 8193 },
  { "message": "Uncaught SyntaxError: Unexpected end of input",  "filename": ".../js/filament/support/support.js",  "lineno": 1, "colno": 8193 },
  { "message": "Uncaught SyntaxError: Invalid or unexpected token", "filename": ".../js/filament/tables/tables.js",   "lineno": 1, "colno": 8189 },
  { "message": "Uncaught SyntaxError: Unexpected end of input",  "filename": ".../build/assets/app.js",            "lineno": 1, "colno": 8001 }
]
```

`colno: 8193` = file received exactly **8192 bytes** (1 × 8KB AMPHP read chunk) before EOF.

Actual file sizes: `livewire.min.js` = 237KB, `support.js` = 140KB, `app.js` = 187KB.

### Root Cause

In `LaravelHttpServer::asset()`, JS files go through this path:

```php
 = fopen('php://temp', 'r+');
 = fread(, filesize());   // reads full file ✓
 = ->rewriteAssetUrl();    // rewrites URLs ✓
fwrite(, );                      // writes to php://temp ✓
rewind();                                // rewinds ✓
return new Response(200, [...], new ReadableResourceStream()); // ← BUG
```

`ReadableResourceStream` from AMPHP sets the stream to **non-blocking mode** via `stream_set_blocking($resource, false)`. For `php://temp` in non-blocking mode, the underlying read returns at most one 8192-byte chunk per read event, and AMPHP's fiber-based event loop does not schedule subsequent reads within the same request lifecycle — so only the first 8KB is ever delivered to Playwright/Chromium.

Non-JS files are also served via `ReadableResourceStream` on the original file handle but they do not cause parse errors (images, CSS, fonts don't fail silently in the same way), so the truncation is hidden.

### Fix

Replace the stream-based response with a plain string response, bypassing async streaming entirely:

```php
private function asset(string $filepath): Response
{
    $rawContent = file_get_contents($filepath);

    if ($rawContent === false) {
        return new Response(404);
    }

    $mimeTypes = new MimeTypes();
    $contentType = $mimeTypes->getMimeTypes(pathinfo($filepath, PATHINFO_EXTENSION));
    $contentType = $contentType[0] ?? 'application/octet-stream';

    if (str_ends_with($filepath, '.js')) {
        $rawContent = $this->rewriteAssetUrl($rawContent);
    }

    return new Response(200, [
        'Content-Type' => $contentType,
        'Content-Length' => (string) strlen($rawContent),
    ], $rawContent);
}
```

This also fixes the 0-byte file crash described in #1635 (no need for the `filesize() === 0` guard in PR #203).

A PR with this fix is available at pestphp/pest-plugin-browser.

### How to Reproduce

Any Laravel + Pest browser test suite that loads JS files larger than 8KB (Livewire, Filament, any Vite bundle). Use `assertNoJavaScriptErrors()` and it will fail. Or capture errors manually:

```php
test('no js errors', function () {
    $page = visit('/');
    $page->page()->waitForLoadState('networkidle');
    $errors = $page->script('() => window.__pestBrowser?.jsErrors ?? []');
    // colno ~8193 on large JS files
    dump($errors);
});
```

### Pest Version

4.3.1

### PHP Version

8.5 (also reproducible on 8.4 / 8.3)

### Operation System

Linux

### Notes

Related to #1635 (0-byte file crash). This is a separate truncation bug affecting all files >8KB.

G4Zz0L1:
## Problem

When running browser tests, all static JS files **larger than 8192 bytes** are truncated to exactly 8KB, causing `SyntaxError: Unexpected end of input` in Playwright. This makes `assertNoJavaScriptErrors()` fail on any page loading Livewire, Filament, or any Vite bundle.

### Evidence

Capturing error details via `window.__pestBrowser.jsErrors` after `waitForLoadState('networkidle')`:

```json
[
  { "message": "Uncaught SyntaxError: Unexpected end of input", "filename": ".../livewire.min.js", "lineno": 1, "colno": 8193 },
  { "message": "Uncaught SyntaxError: Unexpected end of input", "filename": ".../support.js",     "lineno": 1, "colno": 8193 },
  { "message": "Uncaught SyntaxError: Invalid or unexpected token", "filename": ".../tables.js", "lineno": 1, "colno": 8189 }
]
```

`colno: 8193` = exactly **8192 bytes** received — one AMPHP read chunk.

### Root Cause

`ReadableResourceStream` sets the stream to **non-blocking mode** (`stream_set_blocking($resource, false)`). For `php://temp` in non-blocking mode, AMPHP's fiber event loop only delivers the first 8192-byte chunk and does not schedule further reads within the same request, so files >8KB are silently truncated.

## Fix

Replace the `fopen`/`fread`/`php://temp`/`ReadableResourceStream` pipeline with `file_get_contents` and a plain string response body. The AMPHP `Response` accepts a string body and sends it in full, bypassing async read-chunk limitations entirely.

This also supersedes PR #203: the `filesize() === 0` guard is no longer needed since `file_get_contents` on an empty file returns `''` rather than crashing.

Fixes pestphp/pest#1664
Also fixes pestphp/pest#1635

### DIFF
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace Pest\Browser\Drivers;
 
+use Amp\ByteStream\ReadableResourceStream;
 use Amp\Http\Cookie\RequestCookie;
 use Amp\Http\Server\DefaultErrorHandler;
 use Amp\Http\Server\HttpServer as AmpHttpServer;
@@ -316,25 +317,38 @@ final class LaravelHttpServer implements HttpServer
      */
     private function asset(string $filepath): Response
     {
-        $rawContent = file_get_contents($filepath);
-
-        if ($rawContent === false) {
-            return new Response(404);
-        }
-
         $mimeTypes = new MimeTypes();
         $contentType = $mimeTypes->getMimeTypes(pathinfo($filepath, PATHINFO_EXTENSION));
-
         $contentType = $contentType[0] ?? 'application/octet-stream';
 
         if (str_ends_with($filepath, '.js')) {
-            $rawContent = $this->rewriteAssetUrl($rawContent);
+            // ReadableResourceStream sets php://temp to non-blocking mode, causing
+            // AMPHP's event loop to deliver only the first 8192-byte chunk for files
+            // written to a php://temp stream. Using file_get_contents + a string body
+            // sends the full content in one shot, avoiding the truncation.
+            $content = file_get_contents($filepath);
+
+            if ($content === false) {
+                return new Response(404);
+            }
+
+            $content = $this->rewriteAssetUrl($content);
+
+            return new Response(200, [
+                'Content-Type' => $contentType,
+                'Content-Length' => (string) strlen($content),
+            ], $content);
+        }
+
+        $file = fopen($filepath, 'r');
+
+        if ($file === false) {
+            return new Response(404);
         }
 
         return new Response(200, [
             'Content-Type' => $contentType,
-            'Content-Length' => (string) strlen($rawContent),
-        ], $rawContent);
+        ], new ReadableResourceStream($file));
     }
 
     /**
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Pest\Browser\Drivers;
 
-use Amp\ByteStream\ReadableResourceStream;
 use Amp\Http\Cookie\RequestCookie;
 use Amp\Http\Server\DefaultErrorHandler;
 use Amp\Http\Server\HttpServer as AmpHttpServer;
@@ -317,9 +316,9 @@ final class LaravelHttpServer implements HttpServer
      */
     private function asset(string $filepath): Response
     {
-        $file = fopen($filepath, 'r');
+        $rawContent = file_get_contents($filepath);
 
-        if ($file === false) {
+        if ($rawContent === false) {
             return new Response(404);
         }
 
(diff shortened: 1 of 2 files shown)


G4Zz0L1:
Addressed both review comments:

**Memory concern (r3057164015):** Fixed — the `file_get_contents` + string body path is now limited exclusively to `.js` files. All other assets (images, fonts, CSS, videos, etc.) keep the original `ReadableResourceStream` path on the real file handle, which is unaffected by the `php://temp` non-blocking truncation.

**Regression test (r3057164073):** Added in `tests/Unit/Drivers/Laravel/LaravelHttpServerTest.php` — creates a JS file slightly over 8192 bytes with a sentinel string at the end and asserts the full response is returned via `assertSee('END_SENTINEL')`. Without the fix this test would fail because the sentinel falls after the first 8KB chunk.

realpoke:
Can this be merged, running my head against this issue too...

mariomka:
Please merge it. I spent a lot of time investigating what happened.

G4Zz0L1:
Any news on this?

PR #209, opened 2026-04-17
docs: Improve PHPDoc for click() method to reflect smart selector behavior

dmitritarasov:
Add new clickSelector() method to InteractsWithElements trait

Allows clicking on elements by directly passing a data-test selector string


nunomaduro:
what's is the difference between this and `click`?

dmitritarasov:
The current PHPDoc says 'link with text' which is misleading. Let's just update it to reflect the actual smart selector behavior. Then clickSelector() is not needed. 

Updated the PR — replaced the duplicate method with an improved PHPDoc for click() instead. Please take another look.

dmitritarasov:
I've updated the PR title to reflect the final changes.

PR #210, opened 2026-04-22
feat/bugfix: allow-browser-screenshot-config

Perturbatio:
Currently Pest browser snapshot testing is a boolean check, it either precisely matches or it doesn't (via `expect($actualImageBlob)->toMatchSnapshot();` )

This PR adds the ability to control the granularity of screenshot diffing and only fail if it exceeds the settings.

You can retain existing functionality with:
```PHP
   // in Pest.php
    pest()->browser()->screenshotMaxDiffPixelRatio(0)
        ->screenshotMaxDiffPixels(0)
        ->screenshotThreshold(0);
```

The above will fail the test even if a single pixel is off (within Playwright's own internal tolerances).

If you want a *little* tolerance to account for some pixels not rendering consistently, you can do something like:

```PHP
  // in Pest.php
  pest()->browser()
      ->screenshotMaxDiffPixelRatio(0.01)
      ->screenshotMaxDiffPixels(5)
      ->screenshotThreshold(0.3)
```

Larger numbers = more tolerance, the ratio is in the range of 0 to 1 and internally is a formula that determines the number of pixels allowed to vary by multiplying the expected width * height * maxDiffPixelRatio (which *can* result in very large tolerances)

Playwright will prefer the lower of the `maxDiffPixelRatio` or the `maxDiffPixels`

* added ability to control the screenshot diffing via `screenshotThreshold`, `screenshotMaxDiffPixels` and `screenshotMaxDiffPixelRatio`
* fixed issue with Pest assuming that if a screenshot does not precisely match (via base64 string comparison) that it has failed testing


Perturbatio:
Note that other configuration options would be nice to have but I'm focussing on the ones which are most likely to affect most users atm.

Additionally, you could say this is also a bug fix for the way Pest currently does screenshot diffing

PR #211, opened 2026-04-23
fix: give the playwright npm server time to shut down browser children

jordanmiguel:
## The bug

When running a suite with `--headed`, the `pest` process hangs indefinitely after the "Tests: N passed" summary prints. On CI / in shells with a pipe it keeps the shell prompt busy; locally the user has to hit Ctrl-C to reclaim the terminal.

Repro (macOS + Linux, any project with browser tests):

```bash
./vendor/bin/pest tests/Browser --headed
# Summary prints, then hangs.
```

Headless mode is unaffected.

## Root cause

In `PlaywrightNpmServer::stop()`:

```php
$this->systemProcess->stop(
    timeout: 0.1,
    signal: PHP_OS_FAMILY === 'Windows' ? null : SIGTERM,
);
```

Symfony's `Process::stop($timeout)` sends `SIGTERM`, waits `$timeout` seconds, then escalates to `SIGKILL`. 100ms is too short for headed Chromium — its graceful close typically takes 700–1200ms because it has to flush IPC to its GPU/renderer children, persist session state, and tear down its windows.

With the old 100ms grace period the server receives `SIGKILL` while Chromium is still in mid-shutdown; its stdin/stdout pipes stay open on orphan descriptors, so the parent PHP process blocks in `proc_close`/`fclose` waiting on them. That is the observable hang.

## Fix

Bump the grace period to `2.0` seconds. Plenty of headroom for headed Chromium on every platform tested (macOS 14 Apple Silicon, Ubuntu 22.04 x86_64) and still bounded so a truly stuck Playwright server still gets `SIGKILL`.

Headless runs complete faster than the grace period, so there is no perceptible overhead — `stop()` returns as soon as Node exits.

## Before / after

```
# Before
./vendor/bin/pest tests/Browser --headed
# ...Tests: 19 passed (80 assertions) in 48.23s
# hangs indefinitely, Ctrl-C required

# After
./vendor/bin/pest tests/Browser --headed
# ...Tests: 19 passed (80 assertions) in 48.23s
# returns to prompt within ~1s of summary
```

## Notes

- No behavioural change for passing headless runs.
- If Chromium genuinely hangs beyond 2s (e.g. deadlock), `SIGKILL` still fires at T+2s — no regression vs. the current escalate-to-SIGKILL behaviour, just pushed out 1.9s.
- Did not add a new test because the hang only reproduces with a spawned headed Chromium subprocess, which this repo's own suite does not exercise. Happy to add one (e.g. `tests/CoreFeatures/HeadedShutdownTest.php` that launches + closes + asserts the parent exits within N seconds) if that fits the project's testing philosophy.

### DIFF
--- a/src/Playwright/Servers/PlaywrightNpmServer.php
+++ b/src/Playwright/Servers/PlaywrightNpmServer.php
@@ -94,8 +94,14 @@ final class PlaywrightNpmServer implements PlaywrightServer
     public function stop(): void
     {
         if ($this->systemProcess instanceof SystemProcess && $this->isRunning()) {
+            // The Playwright node server needs time to clean up its browser
+            // children — especially Chromium in headed mode, which can take
+            // close to a second to shut down gracefully. With the old 0.1s
+            // timeout, SIGTERM raced the browser teardown and Symfony would
+            // escalate to SIGKILL while children were still detaching,
+            // leaving the parent pest process waiting on zombie file handles.
             $this->systemProcess->stop(
-                timeout: 0.1,
+                timeout: 2.0,
                 signal: PHP_OS_FAMILY === 'Windows' ? null : SIGTERM,
             );
         }


ETES-Stuttgart:
I've got the same issue but only when i use Firefox. This PR fixed it 🎊. Thanks.

Kimtran-art:
I hit this same teardown hang on **Firefox** (headless) and can confirm this PR fixes it there too, not only Chromium in headed mode.

Without the change, `./vendor/bin/pest --browser firefox` runs the specs, they pass, but the PHP process never exits: it wedges in `symfony/process` during teardown and leaves an orphaned `playwright run-server` behind. We had been working around it with a wrapper script that watches the JUnit log and SIGKILLs the wedged process group.

With `timeout: 2.0` (this PR), the exact same run exits cleanly (exit code `0`, ~18s) and leaves no orphaned `run-server`. So on Firefox the old `0.1s` timeout was racing the browser-child teardown as well.

Environment: pest-plugin-browser v4.3.1, Playwright 1.59.x, PHP 8.3, macOS. Would be great to get this merged. Thanks @jordanmiguel!

PR #226, opened 2026-06-13
fix: bind HTTP server to an OS-assigned ephemeral port

G4Zz0L1:
Avoids the find-then-rebind TOCTOU in Port::find() that intermittently
  throws 'Address already in use' when many parallel test workers each
  start a LaravelHttpServer. Bind to port 0 and resolve the actual bound
  port from getServers() after start.


G4Zz0L1:
Any news on this?

PR #227, opened 2026-06-14
Avoid retrying side-effecting browser actions

CalorieStar:
## Summary

I had an issue where `click()` was timing out on some pages, even though the click visibly occurred in the browser when using --debug.

This appears to relate to the behaviour described in:

Related to pestphp/pest#1511

## What changed

This PR updates `AwaitableWebpage` so state-changing browser actions are no longer routed through `Execution::waitForExpectation()`.

This includes actions such as:

* `click`
* `press`
* `fill`
* `check`
* `select`
* `drag`
* `attach`
* similar browser actions that may alter page or application state

Assertions and read-style expectations still use the existing retry loop.

## Why

`Execution::waitForExpectation()` retries the callback using a 1000ms Playwright timeout. That behaviour is useful for assertions, because repeatedly checking whether something is true is usually safe.

For example, retrying an assertion such as `assertSee()` is useful because the page may need time to update.

However, this retry behaviour is less suitable for browser actions. Actions such as `click()` and `press()` can change the page, trigger JavaScript, submit forms, start navigation, or update application state.

In these cases, the action may already have succeeded, but Pest may still report `Timeout 1000ms exceeded` if the operation did not complete within the 1000ms attempt window. Retrying the action can then cause further issues because:

* the original element may no longer exist
* navigation may already have started or completed
* the page may now be in a different state
* the same action may be attempted more than once
* form submissions or JavaScript handlers may be triggered more than intended

This PR avoids the pattern where a browser action visibly succeeds, but Pest still fails because the action was handled through the expectation retry loop.

## Behaviour after this change

State-changing browser actions are attempted once and use the configured Playwright/browser timeout.

Assertions continue to be retried until the configured timeout expires.

This preserves the useful retry behaviour for assertions, while avoiding unsafe retries for actions that may alter browser or application state.

## Side-effecting / state-changing actions

A side-effecting action is an operation that changes something.

In browser testing, this means the command does not merely observe the page. It may alter:

* the DOM
* form values
* browser state
* page state
* application state
* server state
* database records
* navigation state

Examples include `click`, `press`, `fill`, `check`, `select`, `drag`, and `attach`.

These actions should be awaited using the configured action timeout, but not retried in the same way as assertions.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5,7 +5,7 @@
   "packages": {
     "": {
       "dependencies": {
-        "playwright": "^1.59.1"
+        "playwright": "^1.60.0"
       }
     },
     "node_modules/fsevents": {
@@ -23,12 +23,12 @@
       }
     },
     "node_modules/playwright": {
-      "version": "1.59.1",
-      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.59.1.tgz",
-      "integrity": "sha512-C8oWjPR3F81yljW9o5OxcWzfh6avkVwDD2VYdwIGqTkl+OGFISgypqzfu7dOe4QNLL2aqcWBmI3PMtLIK233lw==",
+      "version": "1.60.0",
+      "resolved": "https://registry.npmjs.org/playwright/-/playwright-1.60.0.tgz",
+      "integrity": "sha512-hheHdokM8cdqCb0lcE3s+zT4t4W+vvjpGxsZlDnikarzx8tSzMebh3UiFtgqwFwnTnjYQcsyMF8ei2mCO/tpeA==",
       "license": "Apache-2.0",
       "dependencies": {
-        "playwright-core": "1.59.1"
+        "playwright-core": "1.60.0"
       },
       "bin": {
         "playwright": "cli.js"
@@ -41,9 +41,9 @@
       }
     },
     "node_modules/playwright-core": {
-      "version": "1.59.1",
-      "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.59.1.tgz",
-      "integrity": "sha512-HBV/RJg81z5BiiZ9yPzIiClYV/QMsDCKUyogwH9p3MCP6IYjUFu/MActgYAvK0oWyV9NlwM3GLBjADyWgydVyg==",
+      "version": "1.60.0",
+      "resolved": "https://registry.npmjs.org/playwright-core/-/playwright-core-1.60.0.tgz",
+      "integrity": "sha512-9bW6zvX/m0lEbgTKJ6YppOKx8H3VOPBMOCFh2irXFOT4BbHgrx5hPjwJYLT40Lu+4qtD36qKc/Hn56StUW57IA==",
       "license": "Apache-2.0",
       "bin": {
         "playwright-core": "cli.js"
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
   "dependencies": {
-    "playwright": "^1.59.1"
+    "playwright": "^1.60.0"
   }
 }
--- a/src/Api/AwaitableWebpage.php
+++ b/src/Api/AwaitableWebpage.php
@@ -17,6 +17,27 @@ use Throwable;
  */
 final readonly class AwaitableWebpage
 {
+    private const array NON_RETRYABLE_ACTION_METHODS = [
+        'click',
+        'press',
+        'pressAndWaitFor',
+        'keys',
+        'type',
+        'typeSlowly',
+        'fill',
+        'hover',
+        'rightClick',
+        'select',
+        'append',
+        'clear',
+        'radio',
+        'check',
+        'uncheck',
+        'attach',
+        'drag',
+        'withKeyDown',
+    ];
+
     /**
      * Creates a new awaitable webpage instance.
      *
@@ -34,8 +55,6 @@ final readonly class AwaitableWebpage
     }
 
     /**
-     * Awaits for the given method to assert true or fail.
-     *
      * @param  array<int, mixed>  $arguments
      */
     public function __call(string $name, array $arguments): mixed
@@ -43,13 +62,16 @@ final readonly class AwaitableWebpage
         $webpage = new Webpage($this->page, $this->initialUrl);
 
         try {
-            if (
-                in_array($name, $this->nonAwaitableMethods, true)
-                || Playwright::timeout() <= 1000
-            ) {
-                // @phpstan-ignore-next-line
-                $result = $webpage->{$name}(...$arguments);
+            if ($this->shouldRunOnce($name)) {
+                // Actions should be attempted once.
+                // Playwright itself will wait for actionability up to the configured timeout.
+                $result = Playwright::usingTimeout(
+                    Playwright::timeout(),
+                    // @phpstan-ignore-next-line
+                    fn () => $webpage->{$name}(...$arguments),
+                );
             } else {
+                // Assertions/read expectations may be retried until the browser timeout expires.
                 $result = Execution::instance()->waitForExpectation(
                     // @phpstan-ignore-next-line
                     fn () => $webpage->{$name}(...$arguments),
@@ -60,7 +82,7 @@ final readonly class AwaitableWebpage
 
             try {
                 $browserException = BrowserExpectationFailedException::from($this->page, $e);
-            } catch (Throwable) { // @phpstan-ignore-line
+            } catch (Throwable) {
                 throw $e;
             }
 
@@ -69,16 +91,18 @@ final readonly class AwaitableWebpage
 
         ServerManager::instance()->http()->throwLastThrowableIfNeeded();
 
-        return $result === $webpage
-            ? $this
-            : $result;
+        return $result === $webpage ? $this : $result;
     }
 
-    /**
-     * Return the page instance.
-     */
     public function page(): Page
     {
         return $this->page;
+    }
+
+    private function shouldRunOnce(string $name): bool
+    {
+        return in_array($name, $this->nonAwaitableMethods, true)
+            || in_array($name, self::NON_RETRYABLE_ACTION_METHODS, true)
+            || Playwright::timeout() <= 1000;
     }
 }
--- a/src/Api/AwaitableWebpage.php
+++ b/src/Api/AwaitableWebpage.php
@@ -21,11 +21,10 @@ final readonly class AwaitableWebpage
         'click',
         'press',
         'pressAndWaitFor',
-        'keys',
+        'submit',
         'type',
         'typeSlowly',
         'fill',
-        'hover',
         'rightClick',
         'select',
         'append',
@@ -34,8 +33,6 @@ final readonly class AwaitableWebpage
         'check',
         'uncheck',
         'attach',
-        'drag',
-        'withKeyDown',
     ];
 
     /**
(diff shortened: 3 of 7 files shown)


MrBMT:
Thanks for this, the PR here appears to have solved every issue I’d been having with duplicate clicks, form input etc. with the browser plugin and hasn’t introduced any negative side effects to any of my tests.

Now tests that were sometimes flakey because the browser plugin would double click or type text multiple times sometimes are working as expected every time, hopefully this can be merged soon.

DanielGSoftware:
Same here, would be amazing if this PR can be looked at.

DanielGSoftware:
Again thanks for this, it does fix the navigation-hang case. But I'm following up on my previous comment because when I applied this PR to my test-suite I got two deterministic regressions, both on a `->type()` immediately after a click that reveals the field.                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                 

Both pass on stock 4.x and fail (15s timeout on the type) with this PR, reproducibly in isolation. The retry loop was silently compensating for a race: GuessLocator resolves the target eagerly (via ->count()), and if the reveal-click's effect hasn't landed yet, it falls back to getByText(<fieldName>) and hangs. Run-once removes that cushion, so the race surfaces.                                                                                                                                                                                                                                                                      
                                                                                                                                                                                                                                                                                                                            
The underlying issue is really navigation-specific, so disabling retries for every action feels broader than needed. Would an opt-in `noWaitAfter` (per the suggestion in https://github.com/pestphp/pest/issues/1511) be preferable? Adding `?array $options = nul`l to `click()`/`press()` and forwarding to the Playwright locator, leaving the retry loop intact:   
```php                                                                                                                                                                                                                                                                                                          
  ->press('Submit', ['noWaitAfter' => true])                                                                                                                                                                                                                                                                                
  ->assertPathIs('/next-page');                                                                                                                                                                                                                                                                                             
```
                                                                                                                                                                                                                                                                                                                            
In my tests that fixes the hang (the navigating press opts out of the nav-settle wait; assertPathIs is still retried and waits for the URL to commit) without touching the type-after-reveal tests.

DanielGSoftware:
Made a [PR](https://github.com/pestphp/pest-plugin-browser/pull/233) for that if it would be the preferred solution

MrBMT:
> Again thanks for this, it does fix the navigation-hang case. But I'm following up on my previous comment because when I applied this PR to my test-suite I got two deterministic regressions, both on a `->type()` immediately after a click that reveals the field.
> 
> Both pass on stock 4.x and fail (15s timeout on the type) with this PR, reproducibly in isolation. The retry loop was silently compensating for a race: GuessLocator resolves the target eagerly (via ->count()), and if the reveal-click's effect hasn't landed yet, it falls back to getByText() and hangs. Run-once removes that cushion, so the race surfaces.
> 
> The underlying issue is really navigation-specific, so disabling retries for every action feels broader than needed. Would an opt-in `noWaitAfter` (per the suggestion in [pestphp/pest#1511](https://github.com/pestphp/pest/issues/1511)) be preferable? Adding `?array $options = nul`l to `click()`/`press()` and forwarding to the Playwright locator, leaving the retry loop intact:
> 
> ```
>   ->press('Submit', ['noWaitAfter' => true])                                                                                                                                                                                                                                                                                
>   ->assertPathIs('/next-page');                                                                                                                                                                                                                                                                                             
> ```
> 
> In my tests that fixes the hang (the navigating press opts out of the nav-settle wait; assertPathIs is still retried and waits for the URL to commit) without touching the type-after-reveal tests.

The `noWaitAfter` option is already [depreciated](https://playwright.dev/docs/api/class-locator#locator-click) FYI, and in the case of click at least, will default to true in the future.

However I've found this PR fixes more than just clicking - when watching tests in headed mode often I'd find it was double typing, and this PR fixed that too. Sadly the depreciated `noWaitAfter` option is not being set true for fill, so that issue will remain without a PR like this one.

In the case if your test failures, I feel like all you'd need is an `assertSee()` or similar before the the `type()` (or ideally `fill()`, as `type()` is depreciated).

Unfortunately sometimes you have to deal with particularly flakey things like entering text in to CKEditor instances etc. and this PR really helps there.

CalorieStar:
Thank you @MrBMT and @DanielGSoftware for all the feedback and suggestions.

My original assessment of the problem was that browser actions were being executed through `Execution::waitForExpectation()`.

That method retries the complete callback using a short Playwright timeout. This is appropriate for assertions, where repeatedly checking a condition is normally safe, but it is problematic for state-changing actions such as:

* click()
* press()
* fill()
* type()
* check()
* select()

An action could succeed in the browser but fail to return within the individual retry window. Pest would then execute the complete callback again.

This caused behaviour such as:

* a click visibly succeeding but Pest reporting Timeout 1000ms exceeded
* the same button being clicked more than once
* text being entered more than once
* an action being retried after navigation had already started
* subsequent attempts failing because the original element no longer existed

My initial solution was therefore to classify state-changing browser methods as non-retryable and execute each action once using the full configured Playwright timeout.

That fixed the original navigation and duplicate-action problem. However, it also exposed another issue in the existing locator-resolution behaviour.

The existing retry loop was performing two jobs:

1. retrying the browser action
2. indirectly retrying locator discovery

`GuessLocator` resolves a target eagerly. For non-explicit selectors, it checks potential id and name locators using count().

When an action immediately follows a click that reveals a field asynchronously, the field may not exist when `GuessLocator` first runs.

For example:

```php
->click('Show additional fields')
->fill('email', 'person@example.com')
```

The sequence could be:

1. the reveal click succeeds
2. the page starts inserting the new field
3. fill() begins before the field exists
4. GuessLocator checks the field's id and name
5. neither currently matches
6. it falls back to an exact visible-text locator
7. Playwright waits on the wrong locator and eventually times out

On the original implementation, the outer expectation retry loop reran the entire callback. A later attempt reran `GuessLocator` after the field had appeared, so it selected the correct input.

My initial run-once change removed this accidental locator retry as well as the unsafe action retry. This caused deterministic regressions for actions such as `type()` or `fill()` immediately after revealing a field.

The problem was therefore not that actions should be retried. The problem was that locator resolution and action execution were coupled together.

The noWaitAfter option helps, but this is deprecated by Playwright and only addresses navigation-related clicks, does not address duplicate typing, filling, checking or other actions, leaves the eager locator-resolution race in place and it requires individual tests to know when to opt out of navigation waiting.


The implementation now separates the operation into distinct stages:

1. resolve the correct locator, retrying only this read-only stage when necessary
2. execute the state-changing browser action once
3. continue retrying assertions independently

In other words:

```text
Resolve locator → perform action once → assert the result
```

rather than:

```text
Retry locator resolution + action + result together
```

The generic locator-guessing behaviour has been extended with intent-specific resolvers:

* forClickable()
* forField()
* forCheckable()
* forSelectable()

Each resolver understands what type of element the calling operation requires.

### Clickable controls

`forClickable()` is used for operations such as clicking and pressing buttons.

It can consider appropriate targets such as:

* explicit selectors
* data-test selectors
* IDs
* names
* buttons
* links
* visible text where appropriate

### Editable fields

`forField()` is used for operations such as:

* type()
* typeSlow()
* fill()
* append()
* clear()
* attach()

It considers field-appropriate candidates such as:

* explicit selectors
* data-test selectors
* IDs
* names
* labels
* placeholders

It does not fall back to an unrelated visible text node merely because that text matches the field name.

### Checkable controls

`forCheckable()` is used for:

* radio()
* check()
* uncheck()

It resolves checkbox and radio controls using appropriate selectors, labels and roles. Where a value is supplied, such as:

```php
->radio('contact_method', 'email')
```

the resolver retains the value restriction so the correct control is selected.

### Select controls

forSelectable() is used by select().

It resolves an actual selectable control through explicit selectors, data-test attributes, IDs, names or associated labels, rather than falling back to ordinary page text.

### Retrying locator discovery safely

Locator discovery is read-only, so it can safely be retried while waiting for asynchronously inserted controls.

During each attempt, the resolver checks all suitable candidates and returns when the correct type of element exists.

The state-changing action is not included inside this retry loop.

This means that a delayed field can appear without requiring Pest to repeat the subsequent fill or type operation.

Explicit selectors and data-test selectors remain lazy Playwright locators where possible, allowing Playwright's own actionability waiting to operate normally.

### Executing actions once

After the correct locator has been resolved, the browser action is executed once using the configured Playwright/browser timeout.

This prevents:

* duplicate clicks;
* duplicate form submissions;
* repeated text entry;
* repeated checkbox or radio changes;
* retrying an action after navigation has begun;
* actions being repeated against a page that has already changed.

Assertions and read-style expectations continue to use `Execution::waitForExpectation()` because retrying an observation is safe and remains necessary for asynchronous page updates.

I have also added Regression tests to cover both the original problem and the regression identified during review.

The tests verify action counts where possible rather than only checking the final page state. This ensures that the fix does not merely reach the correct result after performing an action multiple times.

likemusic:
This PR is going in the right direction, and I can add some measurements that might help the review. I've filed pestphp/pest#1755 with the full analysis; summarising the parts relevant here.

**The 1000ms looks like a regression, not a deliberate retry step.** When the loop was introduced (`1dbf414 feat: livewire::test`), the retry window was a `1` second default, so `usingTimeout(1000, ...)` matched the window exactly. Later `9519b23 chore: improves flakiness` made the window configurable (`Playwright::timeout()`) but left the per-attempt budget pinned at 1000ms:

```diff
-    public function waitForExpectation(callable $callback, int|float $timeout = 1): mixed
+    public function waitForExpectation(callable $callback): mixed
     {
+        $timeout = Playwright::timeout();
-        $end = $start + $timeout;
+        $end = $start + ($timeout / 1_000);
         while (microtime(true) < $end) {
             try {
-                return Playwright::usingTimeout(1000, $callback);
+                return Playwright::usingTimeout(1_000, $callback);
```

That is where `Configuration::timeout()` got decoupled from the attempt budget, which is what produces the `Timeout 1000ms exceeded` messages people report while having configured something else.

**One gap worth closing before merge:** `NON_RETRYABLE_ACTION_METHODS` doesn't include the navigation methods — `navigate`, `refresh`, `back`, `forward`. `Page::goto()` always sends `waitUntil: 'load'`, so on any page that takes longer than 1s to load, every loop attempt is doomed and the navigation only actually happens on the final post-loop call — after the loop has burned the entire configured timeout for nothing.

Measured against a minimal repro (a route with a constant, non-blocking 2s delay; full code in pestphp/pest#1755), a single `navigate()` costs `configured timeout + actual load time`:

| `Configuration::timeout()` | wall-clock for one `navigate()` |
|---|---|
| 3000 ms  | 5.32s / 4.81s / 5.38s    |
| 6000 ms  | 8.91s / 8.55s / 8.74s    |
| 12000 ms | 14.82s / 14.89s / 14.74s |

So with this PR as it stands, `click()` is fixed but `navigate()` still scales with the configured timeout rather than with the page. Adding those four methods to the list would cover it.

On the side-effect half of this PR: `typeSlowly()` reproduces the duplication deterministically (12 chars × 100ms > the 1000ms budget), and the input ends up containing the text several times over:

```
'hello-worlhello-worlhello-worlhello-worlhello-worlhello-worlhello-world!'
```

which matches the duplicate clicks/typing reported here and in pestphp/pest#1511.

Mushoz:
I just wanted to chime in that this has fixed flaky tests here as well. Previously this could be flaky:

`$this->page->press('Next');`

A workaround that I had to use was something like:

`$this->page->page()->locator('.action--next')->click();`

With this PR, the workaround is no longer required.

PR #233, opened 2026-07-02
feat: allow passing options to click() and press()

DanielGSoftware:
## Description

`click()` and `press()` currently call the underlying Playwright locator with no options, so there's no way to pass locator options such as `noWaitAfter`, `timeout`, `position`, or `button`.

This adds an optional `?array $options = null` parameter to both, forwarded straight through to the locator's `click()`:

```php
$page->press('Submit', ['noWaitAfter' => true])
    ->assertPathIs('/next-page');
```

## Motivation

The main driver is `noWaitAfter` for actions that trigger a navigation. Today a `click()`/`press()` that navigates can hang until the timeout on pages whose load doesn't settle quickly (e.g. Livewire/SPA pages that fire a request after the `load` event) — see pestphp/pest#1511. `noWaitAfter` lets the action dispatch without blocking on the navigation to settle; a following `assertPathIs`/`assertSee` (still retried) then confirms the result. This is a narrow, opt-in option rather than a change to the global retry behavior.

## Tests

Added forwarding tests for both `click` and `press` — they pass `['button' => 'right']` and assert a `contextmenu` handler fired, which a plain click would not trigger, proving the option reaches the locator. Existing click/press tests are unchanged.

- `composer lint` (pint + rector), `phpstan`, and 100% type-coverage pass
- browser tests pass

### DIFF
--- a/src/Api/Concerns/InteractsWithElements.php
+++ b/src/Api/Concerns/InteractsWithElements.php
@@ -13,10 +13,12 @@ trait InteractsWithElements
 {
     /**
      * Click the link with the given text.
+     *
+     * @param  array<string, mixed>|null  $options
      */
-    public function click(string $text): self
+    public function click(string $text, ?array $options = null): self
     {
-        $this->guessLocator($text)->click();
+        $this->guessLocator($text)->click($options);
 
         return $this;
     }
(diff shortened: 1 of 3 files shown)


fosron:
Is there a plan to merge this soon? We're having constant problems for not having timeouts set up properly. Thanks!

PR #235, opened 2026-07-18
Add playwright routes support

Jonathan-384481:
This pull request adds routes to pest-plugin-browser with minimal changes to the current code. Use with

`visit('/')->withRoute('**/message', fn (PlaywrightRoute $route) => $route->continue());`

or

`$page->route('**/*', fn (PlaywrightRoute $route) => $route->abort());`

Currently supported methods are continue, abort and fulfill

### DIFF
--- a/src/Api/Concerns/HasRoutes.php
+++ b/src/Api/Concerns/HasRoutes.php
@@ -1,0 +1,31 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Api\Concerns;
+
+use Pest\Browser\Api\Webpage;
+use Pest\Browser\Playwright\Route;
+
+/**
+ * @mixin Webpage
+ */
+trait HasRoutes
+{
+    /**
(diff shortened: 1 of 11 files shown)


KhyFee (commented):
Useful addition — `withRoute` / `route` covering continue/abort/fulfill matches the Playwright surface people reach for first.

Notes from the diff:
1. `setNetworkInterceptionPatterns` with a glob is good; confirm overlapping patterns / multiple `withRoute` calls append rather than replace (tests only show a single pattern).
2. PHPDoc says `callable(Route): bool` but fulfill/abort examples look side-effecting — clarify return-value expectations (ignored vs must return true).
3. `Route::reset()` between tests is important; worth asserting in a test that handlers do not leak across `visit()` calls.
4. Naming `PlaywrightRoute` in userland vs internal `Route` is fine; a short README snippet with the three verbs would help discoverability.

Happy to see this land once multi-route + reset behavior is confirmed.

PR #131, opened 2025-08-25
Feature: allow "remote" Playwright and/or HTTP Servers

ss10sb:
### What:

- New Feature

### Description:

This pull request adds the ability to set HTTP host (`--http-host=...`), HTTP bind address (`--http-bind=...`),
Playwright host (`--playwright-host=...`) and Playwright port (`--playwright-port=...`) options.

I run the Playwright server in a container since I don't run a supported Linux distribution and needed a way
to make it accessible from Pest.  The `--http-host=` was needed for URL rewriting, and`--http-bind=` was needed to
allow the HTTP server to bind on an address accessible to Playwright.

#### Significant Changes:

`Locator::setInputFiles`: modified to be a payload since it needs to be transferred to the "remote" server.

`Support::Persist..` and `Support::TempDirectory`: moved a lot of the persistence logic for both the Playwright and
the HTTP servers to the `Support` classes.

#### Potential Issues:

Much like Parallel testing, these options could potentially break 'headed' mode, using the shell to show the diff and
debug mode.  It depends on how the user sets up their "remote" server(s).  In my case, using containers breaks all
three :)

#### Use:

Playwright in container, Pest on the host:
```shell
$ ./vendor/bin/pest --http-host=host.docker.internal --http-bind=0.0.0.0 --playwright-host=127.0.0.1 --playwright-port=3000
```

Playwright in container, Pest in container (see `/docker.test` for example containers):
```shell
$ ./vendor/bin/pest --http-bind=0.0.0.0 --http-host=pest --playwright-port=3000 --playwright-host=playwright
```

PR #237, opened 2026-07-20
fix: attach() is broken with Playwright 1.61.0 and above

likemusic:
## Problem

`attach()` fails on Playwright 1.61.0 and above:

```
localPaths are not allowed when the client is not local
```

`Locator::setInputFiles()` sends the attached file's *local path*:

```php
$params = ['localPaths' => [$path]];
```

Playwright 1.61.0 added a server-side guard in [`fileUploadUtils.ts`](https://github.com/microsoft/playwright/blob/main/packages/playwright-core/src/server/fileUploadUtils.ts):

```ts
if (localPaths && !frame.attribution.playwright.options.isClientCollocatedWithServer)
  throw new Error('localPaths are not allowed when the client is not local');
```

`isClientCollocatedWithServer: true` is set **only** by `inprocess.ts` and `cli/driver.ts` (stdio). Every server path — `createPlaywright({ isServer: true })` — leaves it falsy. This plugin drives `playwright run-server` over a websocket, so the flag is never set and `localPaths` is rejected **always** — regardless of machine, filesystem or host. It is not a remote-only edge case.

### Reproducing it in this repository

`package.json` pins `^1.59.1`, and 1.59.1/1.60.0 have neither the flag nor the error string — which is why `tests/Browser/Webpage/AttachTest.php` is currently green. Installing the version the caret already permits is enough to break it:

```
npm i -D playwright@1.61.1
vendor/bin/pest tests/Browser/Webpage/AttachTest.php
```

```
FAILED  Tests\Browser\Webpage\AttachTest > it may attach a file to a file input
  localPaths are not allowed when the client is not local
FAILED  Tests\Browser\Webpage\AttachTest > it may attach a file to a file input using an id selector
  localPaths are not allowed when the client is not local

Tests: 2 failed
```

With this patch, both pass again on 1.61.1.

## What this changes

Files are sent **by value** via Playwright's inline `payloads` parameter, which is accepted regardless of how the client is connected — so this fixes the 1.61 guard and works on older versions unchanged.

Two details that matter:

- **`buffer` is base64.** The protocol types it as `binary`, which over the JSON channel means base64 — the same convention this plugin already relies on when *decoding* screenshots in `Support\Screenshot::save()`. Sending raw bytes would corrupt binary uploads.
- **`mimeType` is populated** from the file extension via `Symfony\Component\Mime\MimeTypes` (already a dependency, already used in `LaravelHttpServer`). Omitted, Playwright falls back to `application/octet-stream` and the application sees a different MIME type than the user attached.

Playwright caps inline payloads at 50 MB. Exceeding it previously surfaced as an opaque protocol error; it now throws `FileTooLargeException` naming the file and the limit. Missing or unreadable files throw `FileNotReadableException` rather than failing deeper in the protocol.

The logic lives in a small `Support\FilePayload` value object so it is directly testable, keeping `Locator` thin. `attach()`'s public signature is unchanged.

## Tests

`tests/Unit/Support/FilePayloadTest.php` — 10 tests covering the payload shape (`name`/`mimeType`/`buffer`, never `localPaths`), base64 encoding, MIME guessing and its fallbacks, and both failure modes. These are version-independent, so they hold the fix in place on the currently pinned Playwright.

The important one is **content integrity**: a file containing all 256 byte values plus invalid UTF-8 sequences and trailing whitespace round-trips with a matching SHA-256. That guards the encoding step against the same class of corruption fixed in #236.

`tests/Browser/Webpage/AttachTest.php` is unchanged and passes on both 1.59.1 and 1.61.1.

## Note: the Playwright pin

I have deliberately **not** touched `package.json` or `package-lock.json` here, but it is worth flagging: the pin is already `^1.59.1`, which admits 1.61.x. Only `package-lock.json` holds this at 1.59.1, and CI uses `npm ci` — so CI will keep passing either way and will *not* exercise the failing case above. Any contributor running a plain `npm i`, or a Dependabot lock bump, hits the break.

If you would like CI to actually cover this, the lockfile needs bumping to 1.61.x — happy to add that commit, but I did not want to fold a dependency decision into a bugfix PR.

## Note on scope — this is one of two blockers

Getting an upload all the way into a Laravel app needs a second, independent fix: `LaravelHttpServer::handleRequest()` parses only `application/x-www-form-urlencoded` and passes `[] // @TODO files...` to `Request::create()`, so a `multipart/form-data` submission arrives with neither fields nor files. **That half is already solved by #200** (approved, widely confirmed) — this PR deliberately does not duplicate it, and the two touch disjoint files (`Locator.php` vs `LaravelHttpServer.php`).

I verified the pair locally by merging this branch onto #200 and asserting that an attached file reaches `$request->file()` with matching name, size and SHA-256, alongside a sibling text field. It passes with both and fails with either alone. That end-to-end test is not included here because it would fail on `4.x` until #200 lands — happy to contribute it to #200 instead.

Related: #131 also replaces `localPaths` with a buffer, but as a side-note inside a larger feature (currently conflicting) and without base64 encoding or a MIME type. #232 (multi-file `attach()`) is orthogonal and composes with this.

### DIFF
--- a/src/Exceptions/FileNotReadableException.php
+++ b/src/Exceptions/FileNotReadableException.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Exceptions;
+
+use RuntimeException;
+
+/**
+ * @internal
+ */
+final class FileNotReadableException extends RuntimeException
+{
+    //
+}
--- a/src/Exceptions/FileTooLargeException.php
+++ b/src/Exceptions/FileTooLargeException.php
@@ -1,0 +1,15 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Exceptions;
+
+use RuntimeException;
+
+/**
+ * @internal
+ */
+final class FileTooLargeException extends RuntimeException
+{
+    //
+}
--- a/src/Playwright/Locator.php
+++ b/src/Playwright/Locator.php
@@ -6,6 +6,7 @@ namespace Pest\Browser\Playwright;
 
 use Generator;
 use Pest\Browser\Playwright\Concerns\InteractsWithPlaywright;
+use Pest\Browser\Support\FilePayload;
 use Pest\Browser\Support\Selector;
 use RuntimeException;
 
@@ -651,10 +652,13 @@ final readonly class Locator
 
     /**
      * Set input files for a file input element.
+     *
+     * The file is sent by value rather than by path, as Playwright 1.61.0
+     * and above reject `localPaths` for websocket clients.
      */
     public function setInputFiles(string $path): void
     {
-        $params = ['localPaths' => [$path]];
+        $params = ['payloads' => [FilePayload::fromPath($path)->toArray()]];
         $response = $this->sendMessage('setInputFiles', $params);
 
         $this->processVoidResponse($response);
--- a/src/Support/FilePayload.php
+++ b/src/Support/FilePayload.php
@@ -1,0 +1,106 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Pest\Browser\Support;
+
+use Pest\Browser\Exceptions\FileNotReadableException;
+use Pest\Browser\Exceptions\FileTooLargeException;
+use Symfony\Component\Mime\MimeTypes;
+
+/**
+ * A file to be uploaded, expressed as an inline Playwright payload.
+ *
+ * Playwright 1.61.0 and above refuse `localPaths` unless the client is
+ * collocated with the server, a flag only set by the in-process and stdio
+ * drivers — never by `run-server`. As this plugin always connects over a
+ * websocket, files are sent by value instead of by path.
+ *
+ * @internal
+ */
+final readonly class FilePayload
+{
+    /**
+     * The largest buffer Playwright accepts for an inline payload.
+     */
+    private const int MAX_SIZE_IN_BYTES = 50 * 1024 * 1024;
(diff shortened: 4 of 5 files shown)


likemusic:
Closing: this is already fixed on `5.x`, and `4.x` is dormant.

`setInputFiles()` on `5.x` sends the file by value, exactly as proposed here — see 3bcf4964067238533a45f60db95acc170e27a936 (landed 18 July 2026 inside "chore: latest version of `playwright`", shipped in v5.0.0):

```php
$params = ['payloads' => [[
    'name' => basename($path),
    'mimeType' => $mimeType === false ? 'application/octet-stream' : $mimeType,
    'buffer' => base64_encode($contents),
]]];
```

So the client-side half of the upload problem is solved upstream. Recording the rest for anyone who lands here from a broken `->attach()`:

- The reason `localPaths` stopped working is Playwright 1.61.0, which refuses them unless `isClientCollocatedWithServer` is set — a flag only set by the in-process and stdio drivers, never by `run-server`. Since this plugin always connects over a websocket, `localPaths` are rejected regardless of the machine or filesystem. Pinning Playwright below 1.61 is the other way out on `4.x`; note `package.json` allows `^1.59.1`, so only `package-lock.json` keeps CI on 1.59.1.
- Fixing the Playwright layer alone is not enough: the server side has to parse `multipart/form-data` as well, otherwise the file never reaches `$request->file()`. That half is now fixed on `5.x` by #261, merged 14 September. On `4.x` it is still open as #200 (approved), #223 and #252.

PR #240, opened 2026-07-29
feat: save page source on failed assertions

jeremy-langevin-twc:
This PR adds an opt-in capability to save the full HTML source of the page when a browser assertion fails, alongside the failure screenshot, console logs, and JavaScript errors that are already captured:

```php
// tests/Pest.php
pest()->browser()->source();
```

When enabled and a browser assertion fails, the page source is written to `tests/Browser/Source/` and the failure message gains a line pointing to it, mirroring the existing screenshot line:

```
A screenshot of the page has been saved to [Tests/Browser/Screenshots/...]. The source of the page has been saved to [Tests/Browser/Source/...].
```

**Why**: a screenshot cannot show `data-test` attributes, hidden or collapsed elements, hydration state, or the actual markup a failing selector was matched against. Laravel Dusk stores page source on failure for exactly this reason, so this also closes a Dusk-parity gap.

**Why default-off**: rendered pages can be large, and page HTML can expose PII more structurally than a screenshot does.

**Naming**: `source()` follows the existing terse config grammar (`debug()`, `diff()`, `headed()`) and reuses the term the API already uses for page HTML (`assertSourceHas()`); the directory matches Dusk's `tests/Browser/source` in this repo's capitalization.

**Behavior details**:

- The source dump is skipped in the same cases the failure screenshot is skipped — debug mode and screenshot-diff failures — so both artifacts follow one predictable condition.
- Saving the source can never mask the original failure: it is wrapped so that an error in fetching the page content or writing the file leaves the assertion message (including the screenshot and console-log lines) intact.
- `Source::cleanup()` runs at suite boot exactly like `Screenshot::cleanup()`, and `tests/Browser/Source` is gitignored like the screenshots directory.
- Zero behavior change for anyone who doesn't call `source()`.

This is a backwards-compatible, opt-in addition (SemVer minor). Covered by unit tests for the support class and browser tests for both the enabled and disabled failure paths; `composer test` passes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ coverage.xml
 # Playwright
 node_modules/
 /tests/Browser/Screenshots
+/tests/Browser/Source
 
 # MacOS
 .DS_Store
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -124,4 +124,14 @@ final readonly class Configuration
 
         return $this;
     }
+
+    /**
+     * Saves the page source when a browser assertion fails.
+     */
+    public function source(): self
+    {
+        Playwright::setShouldSaveSourceOnFailedAssertions();
+
+        return $this;
+    }
 }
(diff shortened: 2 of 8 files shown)


jeremy-langevin-twc:
Closing this is not ready

PR #241, opened 2026-07-30
fix: bound Playwright requests so a lost or unanswered response cannot hang the suite

### LINKED ISSUE #1781: [Bug]: Browser plugin — actions on a zero-match locator hang forever instead of failing
### What happened?

In `pest-plugin-browser`, an **action** on a locator that matches zero elements never returns and never throws — the test hangs indefinitely. **Assertions** on missing content fail normally, so the contrast is stark: a typo in an `assertSee` gives you a clean failure with a screenshot, while the same typo in a `click` hangs the suite with no output at all.

Filed here because issues are disabled on `pestphp/pest-plugin-browser`.

Practically, this makes a mistyped selector expensive to diagnose — the suite simply stops, with no indication of which line is responsible. It also means a browser suite can't be relied on to terminate: there's no failure mode, only a wedged process.

### How to reproduce

```php
it('hangs on a zero-match click', function () {
    $page = visit('/');

    fwrite(STDERR, "about to click\n");
    $page->click('button[data-nonexistent-hook="definitely-not-here"]');
    fwrite(STDERR, "click returned\n");   // never printed
});
```

`about to click` prints; `click returned` never does. Killed at 90s. Swapping the `click` for `assertSee('definitely not on this page')` fails properly in a few seconds.

The original discovery was the same thing via the text-locator path, which is how a realistic version of this arises. The rendered label was `+ Have a code?`, and `click('Have a code?')` goes through `GuessLocator` to `getByText($text, exact: true)` — matching nothing. Worth noting the obvious "fix" of `click('+ Have a code?')` also fails, differently: `Selector::isExplicit()` treats `+` as a CSS combinator, so the string is routed to `page->locator()` as a selector rather than as text. Neither spelling is easy to spot from a hang.

### What we found while diagnosing

- The PHP process sits 100% idle in `stream_select` inside a fiber, waiting on the Playwright websocket. Sampled with macOS `sample` — **no timer pending**.
- All page HTTP requests were served and closed; browser and renderer processes alive and idle. Nothing stuck on the app side.
- `Client::execute()` does put a deadline in the message — `'params' => ['timeout' => $this->timeout, ...$params]` at `src/Playwright/Client.php:83`, with `$timeout = 5_000` — so a `timeout: 5000` is sent with the `click`. No timeout error ever comes back over the socket, and the `while (true)` loop in `execute()` blocks in `fetch()` forever waiting for a frame that never arrives.
- Assertions are bounded independently of that, via `Execution::waitForExpectation`, which is presumably why they behave correctly while actions don't.

So the deadline appears to be sent but not honoured for action methods. We didn't establish whether the server ignores the param for these calls or the error frame is never emitted — that part is a guess, and we'd rather leave it to you.

### Suggested behaviour

A zero-match action should raise the same kind of error an assertion does, naming the selector, ideally after the configured timeout. Even a crude client-side deadline around action commands would turn this from "the suite is wedged" into "line 42 has a bad selector".

### Package versions

- `pestphp/pest-plugin-browser` 5.0
- `pestphp/pest` 5.0.1
- Laravel 13.23, PHP 8.3.30
- Chromium via `npx playwright install chromium`, macOS 15 (Darwin 25.5.0)

Happy to supply the `sample` output or a minimal repo if useful.

### LINKED ISSUE #1801: [Bug]: Browser plugin — Client::execute() spins at 100% CPU forever when the Playwright connection closes
### What Happened

If the WebSocket connection to the Playwright server closes while `Pest\Browser\Playwright\Client::execute()` is waiting for a response, the test process never fails, never times out and never exits. It busy-loops at ~100% CPU indefinitely. No error, no stack trace, no output — CI runs until the job limit.

Filed here because issues are disabled on `pestphp/pest-plugin-browser`.

The cause is two lines in `src/Playwright/Client.php`. `fetch()` turns a closed connection into an empty string:

```php
private function fetch(WebsocketConnection $client): string
{
    return (string) $client->receive()?->read();   // null -> ''
}
```

and `execute()` loops on it with no bound:

```php
while (true) {
    $responseJson = $this->fetch($this->websocketConnection);
    $response = json_decode($responseJson, true);   // '' -> null

    if (isset($response['error']['error']['message'])) { /* null has no error key */ }

    yield $response;

    if (isset($response['id']) && $response['id'] === $requestId) {
        break;                                      // null has no id key either
    }
}
```

Once the connection is gone, `receive()` returns `null` immediately on every call. `null` becomes `''`, `''` decodes to `null`, `null` has no `error` key so nothing throws and no `id` key so nothing breaks. Because the loop never blocks, no timeout anywhere in the stack can interrupt it.

Instrumenting `fetch()` during a live hang:

```
1785426709.031 EMPTY #1  message=NULL closed=yes closeInfo="Browser closed"
1785426709.032 EMPTY #2  message=NULL closed=yes closeInfo="Browser closed"
...
1785426709.032 EMPTY #15 message=NULL closed=yes closeInfo="Browser closed"
```

Fifteen iterations inside one millisecond, with the connection reported closed every time. `ps` shows state `R` at ~100% CPU for as long as you leave it.

### How to Reproduce

Fresh Laravel 13 app with the React starter kit and `pestphp/pest-plugin-browser`. One browser test, run serially — `--parallel` is not needed:

```php
// tests/Browser/TimeoutTest.php
it('fails within the configured timeout when the text never appears', function (): void {
    visit('/')->assertSee('THIS TEXT NEVER APPEARS ON THE PAGE');
});
```

Start the suite, then close the Playwright server while a request is in flight:

```bash
./vendor/bin/pest --filter='never appears' &
sleep 3
pkill -f "playwright run-server"
```

Expected: the test fails with a connection error.
Actual: Pest never exits. `ps -o %cpu,stat` reports ~100% and state `R` indefinitely.

In the sample repository, `./repro.sh hang 3 --filter='never appears'` does all of this and prints a verdict.

### Sample Repository

https://github.com/joshmanders/pest-plugin-browser-hang-repro

### Pest Version

pest v5.0.2, pest-plugin-browser v5.0.0

### PHP Version

8.5.8

### Operation System

macOS, Linux

### Notes

**Not specific to a Playwright version.** I first hit this on the playwright 1.61.1 → 1.62.0 bump and reported it there (microsoft/playwright#42046). It was closed as out of scope, correctly — the protocol change involved was intentional. Testing the client in isolation shows the spin is not tied to a Playwright version:

| playwright | `Client.php` | connection closes mid-request |
| --- | --- | --- |
| 1.61.1 | as shipped | hang, ~100% CPU, forever |
| 1.62.0 | as shipped | hang, ~100% CPU, forever |
| 1.61.1 | patched | fails in ~2.5s with a clear message |
| 1.62.0 | patched | fails in ~2.5s with a clear message |

1.62.0 did not introduce this; it closed connections more often and exposed something latent. On any version, anything that drops the socket wedges the suite — a crashed browser, an OOM-killed server, a dropped connection on a loaded CI runner.

**Scope of the reproduction.** The steps above force the close with `pkill`, because that isolates the client defect deterministically. I have not reproduced an *organic* close in this minimal app — 60 browser tests under `--parallel` against a healthy server complete cleanly in ~11s. The original occurrence was a 235-test suite under `--parallel`, which hung on macOS and on `ubuntu-latest` in CI. What the forced close demonstrates is that whenever the connection does drop, for any reason, the client cannot survive it.

**Related, but distinct.** #1781 describes an action on a zero-match locator hanging in this same loop, but it blocks idle at 0% CPU waiting for a message that never arrives; this one spins at ~100% CPU on messages that already arrived and can never match. #1754 (the orphaned `run-server`) is a different mechanism, though worth noting it can *look* like this bug when output is piped, since the orphan holds the pipe open and the reader never gets EOF.

**Suggested fix.** Throw when `receive()` returns `null` — a closed connection is an error, not an empty message — and bound the request as a whole rather than per message. I have a PR ready and will link it here.

joshmanders:
A browser suite currently has no failure mode for a request that is never answered — only a wedged process. No error, no timeout, no exit; CI runs until the job limit. This fixes three causes of that, all in `Playwright/Client.php`.

**1. Send the timeout as protocol metadata.** Playwright 1.62.0 (microsoft/playwright#41712) moved the timeout off each method's params onto the protocol-level `metadata`, where an absent value means *no timeout at all*. The client still sent it only in `params`, so every request to a 1.62.0 server was implicitly unbounded. That's why an action on a zero-match locator hangs forever there but fails cleanly on 1.61.1. Sending it in both places is backwards compatible and needs no version check.

Fixes pestphp/pest#1781.

**2. Treat a closed connection as an error.** `fetch()` coerced the `null` that `receive()` returns on a closed connection into `''`, which decodes to `null` — which has neither an `error` key to throw on nor an `id` key to break on. The loop then spins at ~100% CPU forever. This one isn't version specific; it reproduces identically on 1.61.1 and 1.62.0, and anything that drops the socket triggers it — a crashed browser, an OOM-killed server, a flaky runner.

Fixes pestphp/pest#1801.

**3. Bound the request as a whole.** Insurance against future protocol drift. Both bounds are needed and each covers the other's blind spot: the cancellation stops a request the server never answers, where `receive()` blocks forever; the deadline check stops one that keeps receiving unrelated messages, where `receive()` returns without ever suspending and the cancellation never gets scheduled. A per-message timeout bounds neither — messages keep arriving, they're just never the one being awaited.

### Verification

Stock Laravel 13 app, both Playwright versions. Reproduction repo: https://github.com/joshmanders/pest-plugin-browser-hang-repro

| playwright | client | zero-match action | connection closed mid-request |
| --- | --- | --- | --- |
| 1.61.1 | as shipped | `Timeout 5000ms exceeded` 10.8s | hang, 100% CPU, forever |
| 1.62.0 | as shipped | hang, 0% CPU, forever | hang, 100% CPU, forever |
| 1.61.1 | this PR | `Timeout 5000ms exceeded` 11.2s | error in 2.5s |
| 1.62.0 | this PR | `Timeout 5000ms exceeded` 10.6s | error in 2.5s |

The zero-match rows take ~10s rather than the configured 5000ms because the action goes through the assertion retry loop first — that's pestphp/pest#1755, untouched here.

This repo's suite: 359 passed on 1.62.0 under `--parallel`. Four unit tests added; I checked they fail without the fix rather than passing vacuously.

### Related, and deliberately not addressed

pestphp/pest#1759 reports a separate defect in this same loop — the `waitUntil` early-break stranding a command's response and desynchronizing later ones. I've left the break condition exactly as it was so the two changes don't collide; that issue's suggested fix applies cleanly on top of this one.

### Notes for review

- `requestGraceSeconds` is 30s on top of the request timeout, deliberately generous so a slow-but-healthy operation on a loaded CI box doesn't start failing. It's a property rather than a constant only so the tests can shrink it — happy to make it a constant and test it another way.
- `params.timeout` is kept alongside the new `metadata.timeout` so servers older than 1.62.0 keep working.

### Two things found while investigating, not fixed here

Happy to open issues or PRs for either.

- **`waitForLoadState()`, `waitForFunction()` and `waitForURL()` are silent no-ops.** `execute()` is a generator function, so its body doesn't run until iterated — these three call it and discard the result. Confirmed by logging every message written to the socket during `visit('/')->assertNoJavascriptErrors()`, which routes through `waitForLoadState('load')`: zero `waitFor*` messages. Present on `4.x` too. This bears on pestphp/pest#1650 and #1625, which attribute the `withinFrame` hang to `waitForLoadState('networkidle')` never resolving — that call currently does nothing at all, so the hang must originate in the `waitFor(['state' => 'attached'])` that follows it.
- **The persisted server address is trusted without validation.** `ServerManager::playwright()` hands parallel workers `AlreadyStartedPlaywrightServer::fromPersisted()`, which reads `.temp/playwright-server.json` without checking anything is still listening, and `markAsStopped()` only runs on a graceful stop — so a killed run leaves the file pointing at a dead port. Distinct from pestphp/pest#1754, which covers the orphaned process itself.

### DIFF
--- a/src/Playwright/Client.php
+++ b/src/Playwright/Client.php
@@ -81,7 +81,9 @@ final class Client
             'guid' => $guid,
             'method' => $method,
             'params' => ['timeout' => $this->timeout, ...$params],
-            'metadata' => $meta,
+            // Playwright reads the timeout from the metadata since 1.62.0, where an
+            // absent value means no timeout at all. Older servers read it from params.
+            'metadata' => ['timeout' => $this->timeout, ...$meta],
         ]);
 
         $this->websocketConnection->sendText($requestJson);
--- a/src/Playwright/Client.php
+++ b/src/Playwright/Client.php
@@ -4,10 +4,15 @@ declare(strict_types=1);
 
 namespace Pest\Browser\Playwright;
 
+use Amp\Cancellation;
+use Amp\CancelledException;
+use Amp\TimeoutCancellation;
 use Amp\Websocket\Client\WebsocketConnection;
+use Amp\Websocket\WebsocketMessage;
 use Generator;
 use Pest\Browser\Exceptions\PlaywrightOutdatedException;
 use PHPUnit\Framework\ExpectationFailedException;
+use RuntimeException;
 
 use function Amp\Websocket\Client\connect;
 
@@ -30,6 +35,11 @@ final class Client
      * Default timeout for requests in milliseconds.
      */
     private int $timeout = 5_000;
+
+    /**
+     * Seconds allowed on top of the timeout before a request is given up on.
+     */
+    private float $requestGraceSeconds = 30.0;
 
     /**
      * Returns the current client instance.
@@ -88,8 +98,24 @@ final class Client
 
         $this->websocketConnection->sendText($requestJson);
 
+        $allowance = ($this->timeout / 1000) + $this->requestGraceSeconds;
+        $deadline = microtime(true) + $allowance;
+
+        // The cancellation bounds a request the server never answers, and the deadline
+        // bounds one that only ever receives unrelated messages. Neither covers both.
+        $cancellation = new TimeoutCancellation($allowance);
+
         while (true) {
-            $responseJson = $this->fetch($this->websocketConnection);
+            if (microtime(true) > $deadline) {
+                throw $this->unanswered($method, $allowance);
+            }
+
+            try {
+                $responseJson = $this->fetch($this->websocketConnection, $cancellation);
+            } catch (CancelledException) {
+                throw $this->unanswered($method, $allowance);
+            }
+
             /** @var array{id: string|null, params: array{add: string|null}, error: array{error: array{message: string|null}}} $response */
             $response = json_decode($responseJson, true);
 
@@ -131,10 +157,30 @@ final class Client
     }
 
     /**
+     * Builds the failure raised when a request is never answered.
+     */
+    private function unanswered(string $method, float $allowance): RuntimeException
+    {
+        return new RuntimeException(sprintf(
+            'The Playwright server did not answer [%s] within %.1f seconds.',
+            $method,
+            $allowance,
+        ));
+    }
+
+    /**
      * Fetches the response from the Playwright server.
      */
-    private function fetch(WebsocketConnection $client): string
+    private function fetch(WebsocketConnection $client, Cancellation $cancellation): string
     {
-        return (string) $client->receive()?->read();
+        $message = $client->receive($cancellation);
+
+        // Without this, the null returned once the connection closes casts to an empty
+        // string and leaves the caller waiting on a response that can never arrive.
+        if (! $message instanceof WebsocketMessage) {
+            throw new RuntimeException('The Playwright server closed the connection unexpectedly.');
+        }
+
+        return (string) $message->read($cancellation);
     }
 }
--- a/tests/Unit/Playwright/ClientTest.php
+++ b/tests/Unit/Playwright/ClientTest.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+use Amp\Websocket\Client\WebsocketConnection;
+use Amp\Websocket\WebsocketMessage;
+use Pest\Browser\Playwright\Client;
+
+it('sends the timeout as metadata so the server enforces it', function (): void {
+    $connection = $this->createStub(WebsocketConnection::class);
+
+    $sent = null;
+    $connection->method('sendText')->willReturnCallback(function (string $payload) use (&$sent): void {
+        $sent = json_decode($payload, true);
+    });
+
+    $connection->method('receive')->willReturnCallback(function () use (&$sent): WebsocketMessage {
+        return WebsocketMessage::fromText((string) json_encode(['id' => $sent['id']]));
+    });
+
+    $client = new Client();
+    $client->setTimeout(100);
+
+    new ReflectionProperty(Client::class, 'websocketConnection')->setValue($client, $connection);
+
+    iterator_to_array($client->execute('page@1', 'click'));
+
+    expect($sent['metadata']['timeout'])->toBe(100)
+        ->and($sent['params']['timeout'])->toBe(100);
+});
--- a/tests/Unit/Playwright/ClientTest.php
+++ b/tests/Unit/Playwright/ClientTest.php
@@ -6,6 +6,61 @@ use Amp\Websocket\Client\WebsocketConnection;
 use Amp\Websocket\WebsocketMessage;
 use Pest\Browser\Playwright\Client;
 
+it('reports a closed connection instead of looping on it', function (): void {
+    $connection = $this->createStub(WebsocketConnection::class);
+    $connection->method('receive')->willReturn(null);
+
+    $client = new Client();
+    $client->setTimeout(100);
+
+    new ReflectionProperty(Client::class, 'websocketConnection')->setValue($client, $connection);
+    new ReflectionProperty(Client::class, 'requestGraceSeconds')->setValue($client, 0.2);
+
+    expect(fn (): array => iterator_to_array($client->execute('page@1', 'goto')))
+        ->toThrow(RuntimeException::class, 'The Playwright server closed the connection unexpectedly.');
+});
+
+it('gives up on a request that only ever receives unrelated messages', function (): void {
+    $connection = $this->createStub(WebsocketConnection::class);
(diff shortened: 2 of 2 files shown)


bambamboole:
yeah i somehow also encounter this.

GDanielRG:
same

yeapea:
Most of this still applies, but the second commit doesn't any more.

`9f43da2` landed on `5.x` independently as `e84a76d` on 2026-08-10, eleven days after you pushed. Same change, plus a bit more — it pulls a per-request `$timeout` out of `$params['timeout']` and sends it in both `params` and `metadata`.

It's also the only thing making this PR conflict. `git merge-tree origin/5.x pr/241` reports a single content conflict, and the hunk is nothing but the two versions of that same fix. Cherry-pick `0456514` on its own and `src/Playwright/Client.php` goes in clean; the only leftover is `tests/Unit/Playwright/ClientTest.php`, which collides because `9f43da2` creates that file and `0456514` adds to it.

One thing worth carrying across when you rebase:

    $allowance = ($this->timeout / 1000) + $this->requestGraceSeconds;

`e84a76d` introduced a local `$timeout`, and this line wants it. As it stands the grace budget comes off the global default while the request itself might be running on a longer explicit timeout. Nothing breaks today — `Page.php:521` is the only caller passing its own `timeout` (30000 ms) and it still fits — but it stops fitting for anything above `global + 30 s`.

Last thing, since this PR and #247 both say the two changes don't collide. True for `src/`, not for the tests: both add `tests/Unit/Playwright/ClientTest.php` as a new file, so whoever merges second hits an add/add on it. The eight test names don't overlap, so it's just a matter of keeping both.

joshmanders:
@nunomaduro closing this from lack of activity or acknowledgement.

PR #247, opened 2026-08-10
fix: drop response frames addressed to another request

### LINKED ISSUE #1759: [browser plugin] Client::execute() waitUntil early-break strands responses and desynchronizes subsequent commands
> Filed here because pestphp/pest-plugin-browser has issues disabled.

### Description

`Client::execute()` can leave a command's own response stranded in the websocket buffer, after which **every later command on the connection consumes shifted responses**, producing confusing downstream failures far from the cause.

**Version:** pest-plugin-browser v4.3.1 (pest v4, PHP 8.5)

### Root cause

`src/Playwright/Client.php`, `execute()` — the response loop breaks on *either* the command's own response id *or* a matching `waitUntil` lifecycle event:

```php
if (
    (isset($response['id']) && $response['id'] === $requestId)
    || (isset($params['waitUntil']) && isset($response['params']['add']) && $params['waitUntil'] === $response['params']['add'])
) {
    break;
}
```

`Page::goto()` always sends `waitUntil: 'load'`. When the `load` lifecycle event arrives **before** the goto command's own response (ordering depends on server/browser timing), the loop breaks early and goto's `{id: ...}` response is never consumed.

The next command then reads it. `processResultResponse()` accepts the first message carrying `result.value` **without correlating the response id**:

```php
foreach ($response as $message) {
    if (isset($message['result']['value'])) {
        return JavaScriptSerializer::parseValue($message['result']['value']);
    }
}
```

From that point the stream is desynchronized: each command may receive the previous command's result.

### Observed symptoms

- `Page::consoleLogs(): Return value must be of type array, true returned` — the evaluate consumed a stale boolean from a previous `assertScript`.
- `click()` returns success but the page never receives the click (failure screenshots show the page fully rendered with pre-click state).
- Text assertions fail on pages that never received the preceding action.

In our suite (~220 browser tests) this reproduced **deterministically** on GitHub Actions (sequential run, sqlite, both with `--only-shell` and full chromium) while passing locally — the event ordering is environment-timing dependent, so victims appear/disappear with machine speed.

### Suggested fix

Drop responses whose id belongs to a previous command before processing:

```php
// in execute(), after json_decode:
if (isset($response['id']) && $response['id'] !== $requestId) {
    continue;
}
```

With this guard our previously-failing suite is green; the stranded response from a `waitUntil` early-break is simply discarded by the next command instead of being misinterpreted as its result. (Correlating ids in `processResultResponse()` would work as well.)

bepsvpt:
## Summary

`Client::execute()` breaks its read loop when the `waitUntil` lifecycle event arrives:

```php
if (
    (isset($response['id']) && $response['id'] === $requestId)
    || (isset($params['waitUntil']) && isset($response['params']['add']) && $params['waitUntil'] === $response['params']['add'])
) {
    break;
}
```

`Page::goto()` always sends `waitUntil: 'load'`, and the `load` event usually arrives before goto's own response. So the goto response stays in the websocket buffer and is never read. The next `execute()` call reads it, and two problems happen:

- **Errors go to the wrong request.** The error check throws on any error frame, without checking the frame's `id`. When `afterEach` closes the context while a goto is not settled yet, Playwright answers that goto late with a `TargetClosedError`. The next command that reads the socket gets this error. In practice, the next test's first browser operation fails with `Target page, context or browser has been closed` (and is marked risky, because it fails before any assertion). If nothing reads the socket until `terminate()`, `Browser::close` throws instead, and the run exits 1 even when every test passed. The failing test changes from run to run because it depends on timing.
- **Results go to the wrong request.** A leftover success frame can be read as the next command's result (pestphp/pest#1759).

This PR skips frames whose `id` belongs to another request, before the error check. Event frames have no `id`, so they are not affected. Error frames for the current request still throw. The `waitUntil` break condition is not changed.

Fixes pestphp/pest#1759.

This change is compatible with #241, which fixes hang problems in the same loop and already mentions this issue: "that issue's suggested fix applies cleanly on top of this one".

## Verification

I logged every `sendText` / `receive` with request ids on a real Laravel browser suite (12 tests). Every stale `TargetClosedError` frame came from an abandoned `goto`. Results on that suite:

| Client | Runs | Result |
|---|---|---|
| v5.0.0 as shipped | 7 | 6 failed — one victim per run, across 6 different tests; one run passed 12/12 but exited 1 |
| this PR | 6 | all 12/12 passed, exit 0 |

## Tests

Four unit tests for `Client::execute()`, using a stubbed `WebsocketConnection` (same approach as #241):

- drops a response addressed to another request
- does not fail the current request on an error frame addressed to another one
- still fails the current request on its own error frame
- still yields event frames, which carry no id

The first two fail without the `src/` change. The last two make sure existing behavior stays the same.

## AI assistance

Root-cause tracing, fix, and tests: Claude Code (claude-fable-5, xhigh).

### DIFF
--- a/src/Playwright/Client.php
+++ b/src/Playwright/Client.php
@@ -91,6 +91,15 @@ final class Client
             /** @var array{id: string|null, params: array{add: string|null}, error: array{error: array{message: string|null}}} $response */
             $response = json_decode($responseJson, true);
 
+            // A response addressed to another request was stranded by a generator
+            // abandoned before its final frame was read — goto() breaks on its
+            // waitUntil event, leaving its response unconsumed. Processing it here
+            // would desynchronize the stream, and the TargetClosedError a context
+            // close sends to a still-settling goto would fail an unrelated command.
+            if (isset($response['id']) && $response['id'] !== $requestId) {
+                continue;
+            }
+
             if (isset($response['error']['error']['message'])) {
                 $message = $response['error']['error']['message'];
 
--- a/tests/Unit/Playwright/ClientTest.php
+++ b/tests/Unit/Playwright/ClientTest.php
@@ -1,0 +1,115 @@
+<?php
+
+declare(strict_types=1);
+
+use Amp\Websocket\Client\WebsocketConnection;
+use Amp\Websocket\WebsocketMessage;
+use Pest\Browser\Playwright\Client;
+use PHPUnit\Framework\ExpectationFailedException;
+
+it('drops responses addressed to another request', function (): void {
+    $connection = $this->createStub(WebsocketConnection::class);
+
+    $sent = null;
+    $connection->method('sendText')->willReturnCallback(function (string $payload) use (&$sent): void {
+        $sent = json_decode($payload, true);
+    });
+
+    $frames = [
+        ['id' => 'stranded-request', 'result' => ['value' => 'stale']],
+    ];
+
+    $connection->method('receive')->willReturnCallback(function () use (&$frames, &$sent): WebsocketMessage {
+        $frame = array_shift($frames) ?? ['id' => $sent['id']];
+
+        return WebsocketMessage::fromText((string) json_encode($frame));
+    });
+
+    $client = new Client();
+
+    new ReflectionProperty(Client::class, 'websocketConnection')->setValue($client, $connection);
+
+    $messages = iterator_to_array($client->execute('page@1', 'evaluateExpression'));
+
+    expect($messages)->toHaveCount(1)
+        ->and($messages[0]['id'])->not->toBe('stranded-request');
+});
+
+it('does not fail the current request on an error frame addressed to another one', function (): void {
+    $connection = $this->createStub(WebsocketConnection::class);
+
+    $sent = null;
+    $connection->method('sendText')->willReturnCallback(function (string $payload) use (&$sent): void {
+        $sent = json_decode($payload, true);
+    });
+
+    $frames = [
+        ['id' => 'stranded-goto', 'error' => ['error' => ['message' => 'Target page, context or browser has been closed']]],
+    ];
+
+    $connection->method('receive')->willReturnCallback(function () use (&$frames, &$sent): WebsocketMessage {
+        $frame = array_shift($frames) ?? ['id' => $sent['id']];
+
+        return WebsocketMessage::fromText((string) json_encode($frame));
+    });
+
+    $client = new Client();
+
+    new ReflectionProperty(Client::class, 'websocketConnection')->setValue($client, $connection);
+
+    $messages = iterator_to_array($client->execute('browser@1', 'newContext'));
+
+    expect($messages)->toHaveCount(1)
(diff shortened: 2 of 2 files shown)


Michael-Brooks:
I'm also running into this same issue. I hope this PR gets reviewed soon

babul:
I am also running into this issue.

cyppe:
Same.  Feels like this plugin is a bit forgotten. But hope someone will be allowed to help nuno with through PRs to keep the momentum of this awesome plugin going.

likemusic:
Independent confirmation of the symptom, on a suite unrelated to this repository.

The `Target page, context or browser has been closed` failure described here reproduces on a third-party set of ~130 browser tests running against v5.0.1: it strikes before any assertion has run, and the name of the failing test moves from run to run. That matches your description exactly — a late error frame from one request being read by the next one, so the test that reports it is whichever happened to read the socket next.

One observation that may help whoever reviews this: the failing-test-name drift makes the bug easy to misfile. On that suite it had been sitting in a quarantine group, understood as "flaky application code", because nothing about the symptom points at the transport — the test fails before it asserts anything, so there is no assertion to blame, and re-running moves the failure elsewhere.

A second, unrelated thing that makes this class of bug slower to diagnose than it should be: the original exception does not survive a browser failure. `BrowserExpectationFailedException::from()` returns a new `ExpectationFailedException` and passes `$e->getPrevious()` as its third argument rather than `$e`, so the exception that actually came out of Playwright — and its stack — is dropped. Measured on `5.x`: `getPrevious()` on what reaches the test is `NULL`. The reported line is fine (PHPUnit filters the plugin's frames and lands on the call that threw); what is missing is the original error underneath it. I will open that as its own PR — mentioning it here only because it is part of why a transport-level failure reads as an application one.

Nothing to add on the fix itself — skipping frames whose `id` belongs to another request, before the error check, is the right shape.

*(Edited: an earlier version of this comment claimed the reported line need not be the throwing call. That was wrong — I had measured it inside a checkout of the plugin itself, where its frames are not filtered as vendor code. In a normal project the line is correct; the loss is the original exception, not the location.)*

likemusic:
Follow-up, on how this sits next to #266 — both change the same loop, so I applied them together on top of `a78ede5`.

They apply cleanly in either order and do not conflict. Combined, the loop skips frames whose `id` belongs to another request (this PR) and breaks only on its own response, with the `waitUntil` shortcut gone (#266) — the same rule from both ends rather than two competing ones.

Measured on PHP 8.4.24 / Playwright 1.62.1 / Chromium, Linux:

- all 8 tests from both PRs pass together (19 assertions);
- with the source changes reverted and only the test files kept, 5 of 7 fail — so both sets of unit tests genuinely pin the behaviour;
- full suite with both applied: 347 passed, 27 skipped, 1 failed, and that one failure is an external-URL test that cannot reach `example.com` through my proxy. Nothing regressed.

I also spent a while trying to reproduce the stranded-frame symptom at the browser level — a test that deliberately leaves a navigation unsettled when it ends, followed by a clean test — and could not get it to fail on demand. The unit stub here is the reliable formulation.

PR #248, opened 2026-08-11
Add slow motion support for easy debugging

nicoverbruggen:
Playwright has a `slowMo` launch option that delays each interaction, which is what you want when you are watching a headed run to check that a test does the right things in the right order. 

At Spatie, we're using more AI to generate browser tests. This way we can run individual tests with `--slow-mo` and tests tend to play out like they would with real users, who slow down between clicking on buttons, links and interacting with elements on the website.

Currently, Pest's browser plugin does not expose this Playwright flag, and a hacky workaround is required to get this to work today.

I looked at prior PRs and found #230, which was created for Pest 4. I've ported this PR and improved for Pest 5. 

It adds a `--slow-mo` flag, optionally with a value in milliseconds, and `pest()->browser()->slowMo(250)`. I've credited the original author of that PR in my first commit, but I've made adjustments in the second that make the argument parsing etc. more in line with the rest of the code of the project.

This PR also refuses slow motion under `--parallel`, next to headed mode, screenshot diffing and debug assertions. Slow motion is only really useful when you can watch a run, and headed mode is already refused there, so the combination just makes things slower.

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -86,6 +86,16 @@ final readonly class Configuration
     }
 
     /**
+     * Slows down each browser action by the given number of milliseconds.
+     */
+    public function slowMo(int $milliseconds = Playwright::DEFAULT_SLOW_MO): self
+    {
+        Playwright::setSlowMo($milliseconds);
+
+        return $this;
+    }
+
+    /**
      * Sets the browsers userAgent.
      */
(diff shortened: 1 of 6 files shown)


pelmered:
Thanks for the credit, for improving the solution and pushing further for this! 

This would be a great addition and hope it gets merged soon.

PR #254, opened 2026-09-03
fix: stop the playwright server process and keep body-less responses off keep-alive connections

cyppe:
Two independent server-process bugs that together made browser suites flaky on Linux. Both were traced on a real project (Laravel 13, Debian 13 in a container, Playwright 1.62.1) and are reproduced by the two tests added here; each test fails on `5.x` and passes with the change. Supersedes #169 and #211 (both target `4.x`).

## 1. `playwright run-server` outlives the run

`PlaywrightNpmServer::start()` uses `Process::fromShellCommandline()` without an `exec` prefix. PHP's `proc_open` runs the string as `sh -c "…"`, and dash (Debian's `/bin/sh`) forks `node` as a child instead of replacing itself, so `stop()` only ever signals the shell:

| how the server is started | process tree | after `stop()` |
|---|---|---|
| current (`sh -c ./node_modules/.bin/playwright run-server …`) | `sh` → `node` | `sh` exits, `node` re-parented to PID 1, keeps running with its browsers |
| with `exec ` | `node` is the direct child | `node` exits, browsers exit with it |

Every browser run leaked one server plus a headless Chromium tree; after a day of local runs a host held 65 of them (about 8 GB) and had swapped 50 GB. `#169` fixes the same thing by passing an array command (Symfony adds `exec` itself in that case); this PR keeps the shell command line and adds the prefix on POSIX, which also keeps the relative `./node_modules/.bin/playwright` resolution unchanged. `stop()` now gives node 2 s to close its browsers before Symfony escalates to SIGKILL, which is what #211 asked for.

Test: `tests/Unit/Playwright/Servers/PlaywrightNpmServerTest.php` starts a server through the class, asserts the process it owns is `node …` (not `sh -c …`), and asserts that nothing listening on that port survives `stop()`. Skipped on macOS/Windows because it reads `/proc`.

## 2. Body-less responses corrupt keep-alive connections

`LaravelHttpServer` hands Symfony's headers to amphp unchanged. Symfony strips `Content-Length` from 1xx/204/304 responses, and amphp's `Http1Driver` chunk-encodes every HTTP/1.1 response that has no `Content-Length`, so a `response()->noContent()` goes out as

```
HTTP/1.1 204 No Content
transfer-encoding: chunked

0\r\n\r\n
```

RFC 7230 §3.3 forbids a body (and `Transfer-Encoding`) on 204/304. Chromium treats the response as complete after the headers and returns the socket to its pool with the five stray bytes unread; the next request on that socket fails with `net::ERR_INVALID_HTTP_RESPONSE` (captured with `--log-net-log`). In the real project the casualty was the Livewire runtime script after a `POST …/impression → 204` beacon, so pages silently never booted and only the assertions that depended on JavaScript failed, intermittently, depending on which socket Chromium picked.

`handleRequest()` now sends informational and empty responses with `Content-Length: 0` and no body, which is what every production web server does. amphp itself should special-case 204/304 the way it already special-cases HEAD (reported separately), but the plugin should not depend on that.

Test: `tests/Browser/Server/BodylessResponseTest.php` posts to a 204/304 route with `keepalive` and then loads a script; without the fix the script request fails on the reused connection.

## Checks

- `vendor/bin/pest tests/Unit/Playwright/Servers/PlaywrightNpmServerTest.php tests/Browser/Server/BodylessResponseTest.php`: fail on `5.x`, 7 passed with the change
- full `vendor/bin/pest` on Linux: see comment below
- `phpstan`: no errors; `rector` and `pint` clean for the touched files

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -305,9 +305,22 @@ final class LaravelHttpServer implements HttpServer
             }
         }
 
+        $headers = $response->headers->all();
+
+        if ($response->isInformational() || $response->isEmpty()) {
+            // Symfony strips Content-Length from 1xx/204/304 responses and amphp
+            // chunk-encodes any HTTP/1.1 response without one, which appends a
+            // terminating chunk to a response that must not carry a body. Browsers
+            // consider such a response complete after the headers, so the stray
+            // bytes corrupt the next request on the same keep-alive connection.
+            unset($headers['transfer-encoding']);
+            $headers['content-length'] = ['0'];
+            $content = '';
+        }
+
         return new Response(
             $response->getStatusCode(),
-            $response->headers->all(), // @phpstan-ignore-line
+            $headers, // @phpstan-ignore-line
             $content,
         );
     }
--- a/src/Playwright/Servers/PlaywrightNpmServer.php
+++ b/src/Playwright/Servers/PlaywrightNpmServer.php
@@ -60,7 +60,10 @@ final class PlaywrightNpmServer implements PlaywrightServer
             return;
         }
 
-        $this->systemProcess = SystemProcess::fromShellCommandline(sprintf(
+        // `exec` makes the shell replace itself with node, so stop() signals the
+        // server itself. Without it, PHP's `sh -c` wrapper receives the signal and
+        // node is re-parented to PID 1 together with the browsers it launched.
+        $this->systemProcess = SystemProcess::fromShellCommandline($this->execPrefix().sprintf(
             $this->command,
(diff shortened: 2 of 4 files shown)


cyppe:
Full `vendor/bin/pest` on Linux (Debian 13 container, PHP 8.4.24, Playwright 1.62.1, Chromium headless shell 151): 337 passed, 27 skipped (the macOS/CI-only screenshot cases). `phpstan` clean. The amphp side of issue 2 is reported at https://github.com/amphp/http-server/issues/393.

PR #258, opened 2026-09-10
 [5.x] Fix canonical URL handling for subdomain routing

hafezdivandari:
Rework of #224 for 5.x

### Why

Pest's browser plugin binds its in-process HTTP server to `127.0.0.1:<port>` and, before this fix, pinned Laravel's URL generation to that address,`app.url`, `url()`, `route()` and `asset()` all resolved to `http://127.0.0.1:<port>`. The configured host from `withHost(...)` was only patched onto the `Host` header inside the request handler.                                                                                                                                                                                          

So the two halves disagreed: Laravel *routed* as `app.localhost` (subdomain routes matched) but *generated* every URL as `127.0.0.1`. That breaks Inertia navigation, because every client-side `route()` (Ziggy / Wayfinder) returns a `127.0.0.1` href. Clicking a `<Link>` on a page served from `app.localhost:<port>` is then a cross-origin navigation, which drops out of the SPA into a full page load instead of an Inertia visit.                                                                                                                                                                                                        

### Changes

The fix resolves the canonical host in one place and syncs it into `app.url` and the URL generator's origin/asset origin, so `route()`, `asset()` and the `Host` the framework sees all agree with the host the browser is actually on.

**Subdomain routing bug**
Previously, the configured host (`withHost(...)`) was only applied ad-hoc to the `Host` header inside the request handler, while Laravel's `app.url`, `route()`, and `asset()` still pointed at the bound socket IP (`127.0.0.1`). This left URL generation and subdomain routing inconsistent with the host the browser was actually navigating to.

Now the canonical host is resolved in one place and synced into `app.url` and the URL generator's origin/asset origin/scheme. The framework consistently sees the canonical host (e.g. for subdomain routing) even when the request arrived over a different network host like `127.0.0.1`. Syncing happens on boot, when the host changes via `withHost()` (wired through `Configuration::setHost()`), and per-request as a safety net.

~**Amp 128 KiB body limit bug**~
- Extracted to #260


src/Drivers/LaravelHttpServer.php
```diff
@@ -160,10 +166,41 @@ public function bootstrap(): void
             assert($urlGenerator instanceof UrlGenerator);
 
             $this->setOriginalAssetUrl($urlGenerator->asset(''));
+            $urlGenerator->forceScheme('http');
+        }
+
+        $this->syncCanonicalUrl();
```
MrPunyapal: this still uses url() so .js assets point to 127.0.0.1 while page is on canonical host. should be canonicalUrl() no? check rewriteAssetUrl below, it still replaces with url().

src/Drivers/LaravelHttpServer.php
```diff
@@ -99,7 +100,16 @@ public function start(): void
...
 
-        $this->socket = $server = SocketHttpServer::createForDirectAccess(new NullLogger());
+        // Increase the body size limit beyond Amp's 128-KiB
+        // default to prevent deadlocks when reading large
+        // JSON payloads in POST, PUT, and PATCH requests.
+        $this->socket = $server = SocketHttpServer::createForDirectAccess(
+            logger: $logger = new NullLogger(),
+            httpDriverFactory: new DefaultHttpDriverFactory(
+                logger: $logger,
+                bodySizeLimit: 64 * 1024 * 1024,
```
MrPunyapal: this is unrelated to canonical fix. pls split to separate pr.

src/Drivers/LaravelHttpServer.php
```diff
@@ -224,11 +283,17 @@ private function handleRequest(AmpRequest $request): Response
...
+        // app handles the request.
+        $this->syncCanonicalUrl();
+        $canonicalUrl = $this->canonicalUrl();
+
         $uri = $request->getUri();
         $path = in_array($uri->getPath(), ['', '0'], true) ? '/' : $uri->getPath();
         $query = $uri->getQuery() ?? ''; // @phpstan-ignore-line
         $fullPath = $path.($query !== '' ? '?'.$query : '');
-        $absoluteUrl = mb_rtrim($this->url(), '/').$fullPath;
+        $absoluteUrl = mb_rtrim($canonicalUrl, '/').$fullPath;
```
MrPunyapal: same here, we drop actual request host and use canonical. why not build from request uri?

src/Drivers/LaravelHttpServer.php
```diff
@@ -261,15 +326,13 @@ private function handleRequest(AmpRequest $request): Response
...
-            $symfonyRequest->headers->set('Host', $hostHeader);
-            // Also set SERVER_NAME for Laravel routing
-            $symfonyRequest->server->set('SERVER_NAME', $configuredHost);
-            $symfonyRequest->server->set('HTTP_HOST', $hostHeader);
-        }
+        // Ensure the framework sees the canonical host (e.g. for subdomain routing)
+        // even when the browser arrived via a different network host (e.g. 127.0.0.1).
+        $canonicalHost = $this->canonicalHost();
+        $hostHeader = sprintf('%s:%d', $canonicalHost, $this->port);
+        $symfonyRequest->headers->set('Host', $hostHeader);
```
MrPunyapal: we force global host here for all requests. per-visit withHost() will break for later fetch/xhr after goto restores global. should we keep request host if its not 127.0.0.1?

src/Configuration.php
```diff
@@ -102,6 +103,12 @@ public function withHost(?string $host): self
     {
         Playwright::setHost($host);
 
+        $http = ServerManager::instance()->http();
```
MrPunyapal: this now creates http server early just by calling withHost(). before it had no side effect. can we avoid that?

tests/Browser/Visit/SubdomainTest.php
```diff
@@ -70,18 +70,18 @@
         'host' => request()->getHost(),
     ]);
 
-    // Set global host: test.domain
-    pest()->browser()->withHost('test.domain');
+    // Set global host: test.localhost
+    pest()->browser()->withHost('test.localhost');
```
MrPunyapal, line 78: we change test.domain to test.localhost here. so withHost now needs a host that resolves to 127.0.0.1. thats breaking, pls note it in description/docs.

tests/Browser/Visit/CanonicalUrlTest.php
```diff
@@ -0,0 +1,66 @@
...
+declare(strict_types=1);
+
+use Illuminate\Http\Request;
+use Illuminate\Support\Facades\Route;
+use Pest\Browser\ServerManager;
+
+it('exposes the canonical host on request()->url() when withHost is set', function (): void {
+    Route::domain('app.localhost')->get('/canonical/request-url', fn (Request $request) => $request->url());
+
+    pest()->browser()->withHost('app.localhost');
```
MrPunyapal, line 16: we set global host here but never reset. it leaks to other tests. pls add cleanup.

MrPunyapal (changes requested):
not good to merge as is. canonical fix direction is right but needs split and fixes below. pls check inline comments.

src/Drivers/LaravelHttpServer.php
hafezdivandari: really good question, It was actually a workaround to use `Host` header before this PR. I'll recheck this to see if it is still needed.

src/Configuration.php
hafezdivandari: As far as I remember (6 months old PR), without it we can't configure host on global hooks like this:

```php
pest()->beforeAll(function (): void {
    pest()->browser()->withHost('admin.localhost');
})->in('Browser/Admin');

pest()->beforeAll(function (): void {
    pest()->browser()->withHost('customer.localhost');
})->in('Browser/Customer');
```

I'll recheck this to make sure.

MrPunyapal: good point on beforeAll, checked the flow here.

beforeAll sets Playwright::host(), then first browser test in file runs __markAsBrowserTest() -> bootstrap() -> syncCanonicalUrl(), so that case is covered even without the sync in Configuration. and with the socket null guard, the Configuration sync is a no-op that early anyway, it only creates the server instance + port early.

the case where it does matter is withHost() mid-test after server start, for route() calls before the next request. so lets keep the immediate sync but only when server already exists, no early create. something like a hasHttp() check in ServerManager instead of instance()->http(). want me to push that or you want to handle it?

hafezdivandari: I'm going through your comments one-by-one, I'll push in few minutes, but in the meantime please feel free to push any fix you want to this PR.

MrPunyapal (approved):

PR #261, opened 2026-09-10
feat: hand the application the files of a multipart request

Eliaprz:
## What

The in-process Laravel server builds the request from the raw body and leaves the uploaded files aside (`Request::create(..., [], // @TODO files...`). A form submitted from the browser with `enctype="multipart/form-data"` therefore reaches the application without its files, and a `required` file field answers with a validation error.

This reads a multipart body the way PHP does for a real upload: the fields go into the request parameters (nested names such as `tags[]` included), each file goes into a temporary file wrapped in an `Illuminate\Http\UploadedFile` in test mode, and the temporary files are removed once the kernel has terminated the request.

## Why `Illuminate\Http\UploadedFile` in test mode

Laravel keeps an instance of its own class untouched when it converts the request's files, while a Symfony instance is converted again without the test flag, and `isValid()` then fails on `is_uploaded_file()` since nothing was really uploaded through PHP.

## Test

`tests/Browser/Webpage/UploadTest.php` submits a multipart form with a text field, an array field and a file, and asserts the application received the file with its name, size and content, valid, along with the fields.

## Notes

- Only `multipart/form-data` bodies are parsed; `application/x-www-form-urlencoded` keeps its `parse_str` path.
- Found while writing browser tests for a Laravel application whose first step is a PDF upload; the same fix is applied there as a test-only middleware until this lands.

### DIFF
--- a/src/Drivers/LaravelHttpServer.php
+++ b/src/Drivers/LaravelHttpServer.php
@@ -17,6 +17,7 @@ use Illuminate\Contracts\Debug\ExceptionHandler;
 use Illuminate\Contracts\Http\Kernel as HttpKernel;
 use Illuminate\Foundation\Testing\Concerns\WithoutExceptionHandlingHandler;
 use Illuminate\Http\Request;
+use Illuminate\Http\UploadedFile;
 use Illuminate\Routing\UrlGenerator;
 use Illuminate\Support\Uri;
 use Pest\Browser\Contracts\HttpServer;
@@ -241,8 +242,12 @@ final class LaravelHttpServer implements HttpServer
         $method = mb_strtoupper($request->getMethod());
         $rawBody = (string) $request->getBody();
         $parameters = [];
+        $files = [];
         if ($method !== 'GET' && str_starts_with(mb_strtolower($contentType), 'application/x-www-form-urlencoded')) {
             parse_str($rawBody, $parameters);
+        }
+        if ($method !== 'GET' && preg_match('/^multipart\/form-data;.*boundary="?([^";]+)"?/i', $contentType, $boundary) === 1) {
+            [$parameters, $files] = $this->parseMultipartBody($rawBody, $boundary[1]);
         }
         $cookies = array_map(fn (RequestCookie $cookie): string => urldecode($cookie->getValue()), $request->getCookies());
         $cookies = array_merge($cookies, test()->prepareCookiesForRequest()); // @phpstan-ignore-line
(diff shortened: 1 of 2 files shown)


MrPunyapal (dismissed):

src/Drivers/LaravelHttpServer.php
```diff
@@ -283,6 +288,11 @@ private function handleRequest(AmpRequest $request): Response
             throw $e;
         } finally {
             config(['app.debug' => $debug]);
+
+            // The application stored what it needed: the uploads go, as PHP's temporary files would.
+            foreach ($files as $file) {
+                @unlink($file->getPathname());
```
MrPunyapal: non-blocking: real php removes temp files at shutdown, after terminate. a terminate listener reading the upload would find it gone here. consider moving this loop after terminate().

src/Drivers/LaravelHttpServer.php
```diff
@@ -241,9 +242,13 @@ private function handleRequest(AmpRequest $request): Response
         $method = mb_strtoupper($request->getMethod());
         $rawBody = (string) $request->getBody();
         $parameters = [];
+        $files = [];
         if ($method !== 'GET' && str_starts_with(mb_strtolower($contentType), 'application/x-www-form-urlencoded')) {
             parse_str($rawBody, $parameters);
         }
+        if ($method !== 'GET' && preg_match('/^multipart\/form-data;.*boundary="?([^";]+)"?/i', $contentType, $boundary) === 1) {
+            [$parameters, $files] = $this->parseMultipartBody($rawBody, $boundary[1]);
```
MrPunyapal, line 250: non-blocking: same name twice keeps only last file, and nested names like files[avatar] stay flat instead of nesting as php would. fine for now, maybe one line note so nobody rediscovers it. real uploads over 128 kib will also want #260.

MrPunyapal (approved):
looks good, verified test mode need and move() safety against vendor code. leaving two non-blocking notes, neither holds the merge.

src/Drivers/LaravelHttpServer.php
Eliaprz: Done in 9898de1: the temporary files are removed after `terminate()`, through a small `removeUploads()` helper, also called before an exception from `handle()` is rethrown. Thanks for the review.

src/Drivers/LaravelHttpServer.php
Eliaprz: Noted in the docblock of `parseMultipartBody()` in 9898de1: two parts with the same name keep the last file, and a nested name such as `files[avatar]` stays flat where PHP would nest it. The commit is now signed as well.

